PR #17, opened 2026-03-17
feat: Add application boilerplate and build infrastructure

mattkram:
## Summary

Here, we add a very simple Rust "Hello, world!" binary, in order to establish the minimal application boilerplate. We use `pixi` for development environment reproducibility, and add a CI pipeline that will build and test the application in various modes:

* We build the standalone rust binary via `pixi run build-release` (wraps around `cargo build --release`)
* We build a conda package via `pixi run build-conda` (this uses `rattler-build` and the recipe also invokes some very basic smoke tests to ensure the application returns successfully and emits `Hello, world!` to standard out

The primary goal of this pull request is to establish a minimal build & test workflow that we can extend, irrespective of the actual application functionality.

Jira: [CLI-343](https://anaconda.atlassian.net/browse/CLI-343)

[CLI-343]: https://anaconda.atlassian.net/browse/CLI-343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main.rs
```diff
@@ -0,0 +1,17 @@
+fn greeting() -> &'static str {
+    "Hello, world!"
+}
+
+fn main() {
+    println!("{}", greeting());
```
zfralish, line 6: 🦀

zfralish (approved):

PR #29, opened 2026-03-20
feat: Add ability to self-update in place

mattkram:
## Summary

Adds self-update functionality to `ana` CLI, allowing users to check for and install updates directly from the command line. This PR is deliberately minimal, with the goal only of allowing explicit check and update in-place. The intent is to establish the end-to-end patterns for how we track available versions, and then download the appropriate artifact if an update is possible.

This unlocks future features around update notification, automation, etc. but those are enhancements that are explicitly deferred.

We also use a custom CLI parsing approach, rather than a framework. We will do that separately, and more intentionally.

### Usage

The feature is implemented under the `ana self` subcommand. The interface is subject to change without warning!

- `ana self update` - Downloads and installs the latest version, replacing the running binary in-place. User will be prompted for confirmation.
- `ana self update --yes/-y` - Same as above, but without user confirmation
- `ana self update --check` - Checks if a newer version is available without installing
- `ana self update --list` - Lists all available releases, marking the current version

### Implementation

- Fetches releases from GitHub API via this repo (requires `GITHUB_TOKEN` for private repo access right now). You can use `export GITHUB_TOKEN=$(gh auth token)` if you have the `gh` CLI installed.
- Supports filtering pre-releases via `ANA_PRERELEASES=false` environment variable. Default (currently) is to show pre-releases. We will flip the default once we are more mature. Later, this will be configurable.
- Platform detection for darwin-arm64, linux-x86_64, and windows-x86_64
- Progress bar during download with ETA
- Semantic version comparison for determining update availability

### Added dependencies

This PR looks huge because of `Cargo.lock`, but the actual implementation is quite small. I have added several dependencies, which are all common and will be needed later. Specifically:

- **reqwest** - HTTP client for fetching releases from GitHub API (blocking mode with JSON and rustls-tls) (we may need to go with async later, but for now this is simplest approach)
- **serde** / **serde_json** - JSON deserialization for API responses
- **semver** - Semantic version parsing and comparison
- **self-replace** - Replace the running binary in-place during updates
- **indicatif** - Progress bar for download status
- **indoc** - Multi-line string formatting for help text (compile-time macros only)

### How to test this

There is a bootstrapping issue in that this feature can't be used until a release is available with it, from which subsequent releases can be updated-to. Here is the best way to try it out from this branch:

* Pull down this PR
* Run `SETUPTOOLS_SCM_PRETEND_VERSION=0.0.1 pixi run build-debug`, which will compile to `./target/debug/ana`. The `SETUPTOOLS_SCM_PRETEND_VERSION` is required to mock building an older version.
* If you want, you can put that on `PATH`: `export PATH=./target/debug:$PATH`
* Update the binary: `./target/debug/ana self update`

Jira: [CLI-354](https://anaconda.atlassian.net/browse/CLI-354)

[CLI-354]: https://anaconda.atlassian.net/browse/CLI-354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6,6 +6,7 @@ version = 4
 name = "ana"
 version = "0.0.0"
 dependencies = [
+ "indicatif",
  "indoc",
  "reqwest",
  "self-replace",
@@ -73,6 +74,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
+name = "console"
+version = "0.15.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "054ccb5b10f9f2cbf51eb355ca1d05c2d279ce1804688d0db74b4733a5aeafd8"
+dependencies = [
+ "encode_unicode",
+ "libc",
+ "once_cell",
+ "unicode-width",
+ "windows-sys 0.59.0",
+]
+
+[[package]]
 name = "displaydoc"
 version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


bkreider (approved):
I tested the upgrade functionality and it worked for me.

PR #31, opened 2026-03-20
chore: Add contribution guide

redeux:
## Summary

Contribution guide


CONTRIBUTING.md
```diff
@@ -0,0 +1,94 @@
...
+| `build-conda` | Build the conda package |
+| `test` | Run unit tests |
+| `pre-commit` | Run pre-commit hooks on all files |
+
+For example:
+
+```bash
+pixi run build-debug
+pixi run test
+```
```
mattkram, line 28: We should unify this from `README.md`, and add a link there if we want it in a separate file.

redeux: When we go live I'd like to keep the README.md practitioner focused and include the technical details for contribution in this file. I think it's cleaner that way and speaks to the audience we want to engage with on the main repo page.

mattkram: I agree. What I was trying to say is that I added a section like this to readme already, so need to remove it. I can get to that tomorrow.

mattkram (approved):

PR #44, opened 2026-03-27
feat: Add authentication commands

mattkram:
## Summary

Here, we add the final authentication commands, providing near-functional parity with `anaconda-auth`. We only implement device flow and file-based keyring for now. System keyring integration will come later, but the PKCE flow will not be reproduced in `ana`.

### Dependent PRs

This PR was originally quite large (~1700 lines), but was split instead into several smaller layered PRs, which are designed to be reviewed and merged in this order:

* #47: Adds minimal configuration support
* #49: Adds login command supporting device flow
* #50: Adds API key storage mechanism and commands for retrieval and logout
* This PR: Adds the final `ana whoami` command, including a basic `ApiClient` for making authenticated requests

### Changes

A summary of aggregated changes is below. In totality, the PR sequence provides round-trip support around required authentication commands:

- Add authentication commands: `login`, `logout`, `whoami`, and `auth api-key`
- Implement OAuth 2.0 Device Authorization Grant (RFC 8628) for login flow
- Add keyring storage compatible with anaconda-auth's JSON format
- Add `ApiClient` struct for authenticated HTTP requests
- Support both top-level commands (`ana login`) and namespaced (`ana auth login`)

## Implementation summary

The implementation adds two main functional modules: `config` and `auth`. The layout is:

- **auth/actions.rs**: login/logout/whoami/show_api_key implementations + ApiClient
- **auth/keyring.rs**: JSON keyring with base64-encoded credentials (multi-domain support)
- **auth/api_keys.rs**: API key creation after OAuth flow
- **config.rs**: Added `keyring_path` field (default: `~/.ana/keyring`, override: `ANA_KEYRING_PATH`)
- **cli.rs**: Added `auth` subcommand group and top-level auth commands

The structure supports extension down the road as we add more features, but is deliberately constrained.

## Configuration

The implementation provides defaults, and each field can be overridden by an environment variable. For example, `ANA_AUTH_DOMAIN=anaconda.com` will log into production. We are defaulting to `stage.anaconda.com` for now.

## Command summary

- [ ] `ana login` (or `ana auth login`) completes device flow and saves API key to disk
- [ ] `ana logout` (or `ana auth logout`) removes API key from keyring. Deletes the file if it becomes empty.
- [ ] `ana whoami` (or `ana auth whoami`) displays user info (or helpful message if not logged in)
- [ ] `ana auth api-key` prints API key (or helpful message if not logged in)
- [ ] `ana config` shows the configuration
- [ ] Keyring format compatible with `~/.anaconda/keyring`

Jira: [CLI-350](https://anaconda.atlassian.net/browse/CLI-350)

[CLI-350]: https://anaconda.atlassian.net/browse/CLI-350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/auth/actions.rs
+++ b/src/auth/actions.rs
@@ -1,4 +1,4 @@
-//! Authentication actions (login, logout).
+//! Authentication actions (login, logout, whoami).
 
 use std::thread;
 use std::time::Duration;
@@ -12,11 +12,60 @@ use crate::config::Config;
 
 const REQUEST_TIMEOUT: Duration = Duration::from_secs(30);
 
-/// Build an HTTP client with standard settings.
-fn build_client() -> Result<reqwest::blocking::Client, AuthError> {
-    Ok(reqwest::blocking::Client::builder()
-        .timeout(REQUEST_TIMEOUT)
-        .build()?)
+/// HTTP client with configuration and optional authentication.
+pub struct ApiClient {
+    client: reqwest::blocking::Client,
+    config: Config,
+    api_key: Option<String>,
+}
+
+impl ApiClient {
+    /// Create a new API client, loading credentials from the keyring if available.
+    pub fn new() -> Result<Self, AuthError> {
+        let config = Config::load();
+        let api_key = get_api_key(&config)?;
+        let client = reqwest::blocking::Client::builder()
+            .timeout(REQUEST_TIMEOUT)
+            .build()?;
+
+        Ok(Self {
+            client,
+            config,
+            api_key,
+        })
+    }
+
+    /// Check if the client has valid credentials.
+    pub fn is_authenticated(&self) -> bool {
+        self.api_key.is_some()
+    }
+
+    /// Get the configured domain.
+    pub fn domain(&self) -> &str {
+        &self.config.domain
+    }
+
+    /// Make an authenticated GET request to an API endpoint.
+    pub fn get(&self, path: &str) -> Result<reqwest::blocking::Response, AuthError> {
+        let url = format!("https://{}{}", self.config.domain, path);
+        let mut request = self.client.get(&url);
+
+        if let Some(ref api_key) = self.api_key {
+            request = request.bearer_auth(api_key);
+        }
+
+        Ok(request.send()?)
+    }
+
+    /// Get the underlying HTTP client for custom requests.
+    pub fn raw_client(&self) -> &reqwest::blocking::Client {
+        &self.client
+    }
+
+    /// Get the configuration.
+    pub fn config(&self) -> &Config {
+        &self.config
+    }
 }
 
 /// OpenID Connect discovery document.
@@ -52,8 +101,9 @@ struct TokenErrorResponse {
 
 /// Perform the device authorization flow.
 pub fn login() -> Result<(), AuthError> {
-    let config = Config::load();
-    let client = build_client()?;
+    let api_client = ApiClient::new()?;
+    let client = api_client.raw_client();
+    let config = api_client.config();
 
     // TODO(mattkram): Better handling for common exceptions like SSL cert, etc.
     // Fetch OpenID configuration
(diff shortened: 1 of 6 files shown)


src/config.rs
```diff
@@ -0,0 +1,302 @@
...
+//! Configuration is loaded from environment variables. In the future, this will
+//! also support reading from `~/.ana/config.toml`.
+//!
+//! # Environment Variables
+//!
+//! | Variable             | Default           | Description                    |
+//! |----------------------|-------------------|--------------------------------|
+//! | `ANA_AUTH_DOMAIN`    | `anaconda.com`    | Authentication domain          |
+//! | `ANA_AUTH_CLIENT_ID` | (Anaconda's ID)   | OAuth client ID                |
+//! | `ANA_SSL_VERIFY`     | `true`            | SSL certificate verification   |
```
AlbertDeFusco: should this be `ANA_AUTH_SSL_VERIFY` to match more closely with existing anaconda-auth or are you thinking of differentiated opportunity for the "auth" and "api" portions?

src/config.rs
```diff
@@ -0,0 +1,302 @@
+//! Global configuration for ana.
+//!
+//! Configuration is loaded from environment variables. In the future, this will
+//! also support reading from `~/.ana/config.toml`.
+//!
+//! # Environment Variables
+//!
+//! | Variable             | Default           | Description                    |
+//! |----------------------|-------------------|--------------------------------|
+//! | `ANA_AUTH_DOMAIN`    | `anaconda.com`    | Authentication domain          |
```
AlbertDeFusco: see below I'd be ok with `ANA_DOMAIN`

src/config.rs
mattkram: Good catch, will update.

I don't have too much preference each way, but just think we should be consistent.

Even the names could become the same. I just didn't want to conflict with anaconda-auth right now. Maybe I'm being too cautious?

mattkram: I have updated to use `ANA_DOMAIN` in #47.

src/config.rs
mattkram: See #47

src/auth/actions.rs
```diff
@@ -12,6 +12,62 @@ use crate::config::Config;
...
+    pub fn domain(&self) -> &str {
+        &self.config.domain
+    }
+
+    /// Make an authenticated GET request to an API endpoint.
+    pub fn get(&self, path: &str) -> Result<reqwest::blocking::Response, AuthError> {
+        let url = format!("{}{}", self.config.base_url(), path);
+        let mut request = self.client.get(&url);
+
+        if let Some(ref api_key) = self.api_key {
```
zfralish: I think we should go ahead and setup a request builder in the init. Something like:

```rust

pub fn new() -> Result<Self, AuthError> {
    let config = Config::load();
    let api_key = get_api_key(&config)?;

    let mut default_headers = HeaderMap::new();
    if let Some(ref key) = api_key {
        let mut auth_value = HeaderValue::from_str(&format!("Bearer {}", key))
            .map_err(|_| AuthError::InvalidKey)?;
        auth_value.set_sensitive(true);  // keeps it out of debug logs
        default_headers.insert(header::AUTHORIZATION, auth_value);
    }

    let client = reqwest::blocking::Client::builder()
        .timeout(REQUEST_TIMEOUT)
        .default_headers(default_headers)
        .build()?;

    Ok(Self { client, config, api_key })
}
```

That way our get,post function are really simple

```rust
pub fn get(&self, path: &str) -> Result<reqwest::blocking::Response, AuthError> {
    let url = format!("{}{}", self.config.base_url(), path);
    Ok(self.client.get(&url).send()?)
}

```

src/auth/actions.rs
```diff
@@ -45,10 +101,9 @@ struct TokenErrorResponse {
 
 /// Perform the device authorization flow.
 pub fn login() -> Result<(), AuthError> {
-    let config = Config::load();
-    let client = reqwest::blocking::Client::builder()
-        .timeout(REQUEST_TIMEOUT)
-        .build()?;
+    let api_client = ApiClient::new()?;
+    let client = api_client.raw_client();
```
zfralish: Why do we need to manually initialize these and break them out?

Couldn't we have them initialized as a part of `ApiClient::new()` and then directly access like `api_client.client.do_something()`?

src/cli.rs
zfralish, line 82: Mentioned in another PR but would like to see these long string broken out into antoher file

zfralish (approved):

src/auth/actions.rs
mattkram: Let's do that. I started and then reverted because it felt premature.

Before we do that I think it might be good to go async early.

src/cli.rs
mattkram: Yes. I am deferring this intentionally because I am currently working with UX on what design we want for our help outputs.

Once we know, then we can most likely construct it directly via the clap parser.

src/auth/actions.rs
mattkram: I made this change. Great suggestion. Decided we can defer async until a follow-up PR.

src/auth/actions.rs
mattkram: I ended up just initializing a new client directly. The login case is the exception, since we want a bare unauthenticated client, while ApiClient is more targeted towards hitting API endpoints. We can always revisit later

PR #45, opened 2026-03-28
feat: Add QR code to login for browserless devices

mcg1969:
## Summary

- Embeds the [url2qr](https://github.com/Anaconda-Sandbox/url2qr) library (BSD-3-Clause) as a vendored Rust module (`src/qr.rs`) — zero new external dependencies beyond `console` (already in the dep tree via `indicatif`)
- Adds QR code rendering of the device auth URL during `ana login`, enabling authentication from devices without a browser by scanning the code on a phone
- Adapts the login output based on whether the browser opens successfully or not

## Design considerations

**Two display modes based on browser availability:**

- **Browser opens:** Clean, spacious layout matching the existing UX. A `(press q for QR code)` hint is shown on the "Waiting for authentication..." line. Pressing `q` renders the QR code inline. The keypress is checked every 100ms during the poll sleep for responsive feedback.
- **Browser fails (or `ANA_OPEN_BROWSER=false`):** QR code is shown immediately above the URL in a compact layout to keep the URL and "Waiting..." message visible without scrolling. The messaging reads "scan the QR code or visit:" to direct users to both options.

**QR code rendering:**

- Uses Unicode half-block characters (▀/▄/█) for 2:1 vertical compression — each terminal line represents 2 QR modules
- `invert: true` for white-on-dark rendering (standard dark terminal backgrounds)
- Quiet zone of 1 module (smaller than spec's 4, but works reliably with phone cameras) to save terminal space
- 4-space indent from the left terminal edge
- Auth URLs (~82 chars) fit well within the 134-char limit (QR version 5)

**Vendoring rationale:** url2qr is a small internal library (~500 LOC, zero dependencies) purpose-built for this use case. Embedding avoids a crates.io publish/dependency and keeps the build self-contained. The source file header notes the upstream repo for future syncing.

## Test plan

- [ ] `cargo test` — all 60 tests pass (includes QR module unit tests for GF tables, version selection, matrix generation, finder patterns, terminal output)
- [ ] `cargo run -- login` — browser opens, clean layout shown, press `q` to reveal QR code, scan with phone to verify it resolves to the auth URL
- [ ] `ANA_OPEN_BROWSER=false cargo run -- login` — QR code shown immediately, URL visible below, verify QR scans correctly
- [ ] Test on a terminal with a small window height to verify the URL doesn't scroll off screen in the no-browser case

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,7 @@ dependencies = [
  "base64",
  "clap",
  "comfy-table",
+ "console",
  "dirs",
  "indicatif",
  "indoc",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,6 +13,7 @@ dependencies = [
  "dirs",
  "indicatif",
  "indoc",
+ "libc",
  "reqwest",
  "self-replace",
  "semver",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,9 @@ name = "ana"
 version = "0.0.0" # Placeholder - actual version derived from git tags via build.rs
 edition = "2024"
 
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
+
 [dependencies]
 base64 = "0.22"
 dirs = "6"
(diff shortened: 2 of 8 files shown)


mattkram (approved):
Looks great! I tested it locally with and without browser enabled and it's all working well.

I refactored a bit. Once we implement tokio, some of the nested polling for keystroke listening can be simplified and cleaned up further, but for now I think it's good enough to :shipit:

PR #47, opened 2026-03-30
feat: Add minimal configuration system and `ana config` command to show

mattkram:
## Summary

Adds a minimal configuration system for `ana` with environment variable support and a CLI command to view current settings.

- Add `console.rs` module with table formatting utility function for consistent CLI output
- Add `config.rs` module that loads configuration from environment variables
- Wire up `ana config` command to display current configuration in a formatted table
- Add `temp-env` dev dependency for testing environment variable handling

> [!NOTE]
> This PR was extracted from #44 

## Configuration Options

| Variable | Default | Description |
|----------|---------|-------------|
| `ANA_DOMAIN` | `stage.anaconda.com` | Authentication domain |
| `ANA_AUTH_CLIENT_ID` | (Anaconda's ID) | OAuth public client ID |
| `ANA_SSL_VERIFY` | `true` | SSL certificate verification |
| `ANA_OPEN_BROWSER` | `true` | Auto-open browser during login |

> [!NOTE]
> The `ANA_DOMAIN` will default to `anaconda.com` before first public release

## Test Plan

- [x] Unit tests for config loading and boolean parsing
- [x] Unit tests for table formatting
- [x] Integration tests for `ana config` command output
- [x] All Rust unit tests pass
- [x] All Python integration tests pass

### Demo (image)

image
image

Jira: [CLI-372](https://anaconda.atlassian.net/browse/CLI-372)

[CLI-372]: https://anaconda.atlassian.net/browse/CLI-372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/cli.rs
zfralish, line 71: For the future, these might be betters in a "constants.rs" file or something similar.

src/config.rs
```diff
@@ -0,0 +1,239 @@
...
+        ];
+        println!(
+            "{}",
+            crate::console::format_table(("Setting", "Value"), &rows)
+        );
+    }
+}
+
+/// Convert a boolean to a string.
+fn bool_to_str(val: bool) -> &'static str {
```
zfralish, line 86: Maybe pull all these into a `utils.rs` file

src/config.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    #[test]
+    fn test_parse_bool_env_nonexistent_returns_default() {
+        // When not set, returns default
+        assert!(parse_bool_env("NONEXISTENT_VAR_12345", true));
+        assert!(!parse_bool_env("NONEXISTENT_VAR_12345", false));
+    }
+
+    #[test]
```
zfralish, line 131: Do we want tests in the same file?

src/console.rs
```diff
@@ -0,0 +1,97 @@
...
+
+// ANSI escape codes
+const BOLD: &str = "\x1b[1m";
+const RESET: &str = "\x1b[0m";
+
+/// Format rows as a unicode table with headers.
+///
+/// Each row is a (key, value) pair. The table auto-sizes columns to fit content.
+/// Headers are displayed in bold.
+pub fn format_table(headers: (&str, &str), rows: &[(&str, &str)]) -> String {
```
zfralish: I would prefer we find a lib for this sort of stuff, I think this sort of UI thing becomes a nightmare was we expand capability.

zfralish (approved):

src/console.rs
mattkram: I agree. I'll pull in comfy-table, as it's already widely used in our ecosystem.

src/config.rs
mattkram: For now, I recommend we keep it here unless it's needed elsewhere. I'm not a big "box of utils" fan personally, unless it really makes sense.

src/cli.rs
mattkram: Agree. For now, let's leave it until we find need for more global constants. That'll be an easy future refactor.

src/config.rs
mattkram: Yes, this is actually the recommended way in the [Rust book](https://doc.rust-lang.org/book/ch11-03-test-organization.html). One of the main reasons (for unit tests) is you can access private functions for testing, without exposing through the `pub` interface.

PR #52, opened 2026-03-31
feat: Add telemetry for CLI command execution

zfralish:
## Summary
- Add `anaconda-otel-rs` dependency for OpenTelemetry metrics
- Implement telemetry middleware pattern on `Action` with `execute()` method
- Track command invocations, successes, and failures via `increment_counter`

## Changes
- `Action::execute()` wraps command execution with telemetry
- `Action::name()` returns command identifier for metrics
- Simplify `main()` to single `cli::parse().execute()` call
- Initialize telemetry on startup with metrics export to Anaconda Connect

## Test plan
- [x] All existing tests pass
- [ ] Verify metrics appear in telemetry backend with `OTEL_API_KEY` set

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -5,6 +5,7 @@ use clap::{Parser, Subcommand};
 use indoc::formatdoc;
 
 use crate::VERSION;
+use crate::auth;
 use crate::config::{self, Config};
 use crate::update;
 
@@ -38,8 +39,13 @@ impl Action {
         match self {
             Action::ShowHelp => "help",
             Action::ShowSelfHelp => "self.help",
+            Action::ShowAuthHelp => "auth.help",
             Action::ShowVersion => "version",
             Action::ShowConfig => "config",
+            Action::Login => "login",
+            Action::Logout => "logout",
+            Action::ShowApiKey => "auth.api-key",
+            Action::Whoami => "whoami",
             Action::Update { .. } => "self.update",
             Action::CheckForUpdate => "self.update.check",
             Action::ShowAvailableVersions => "self.update.list",
@@ -77,10 +83,22 @@ impl Action {
                 print_self_help();
                 Ok(())
             }
+            Action::ShowAuthHelp => {
+                print_auth_help();
+                Ok(())
+            }
             Action::ShowVersion => {
                 println!("{}", VERSION);
                 Ok(())
             }
+            Action::ShowConfig => {
+                Config::load().print_table();
+                Ok(())
+            }
+            Action::Login => Ok(auth::login()?),
+            Action::Logout => Ok(auth::logout()?),
+            Action::ShowApiKey => Ok(auth::show_api_key()?),
+            Action::Whoami => Ok(auth::whoami()?),
             Action::Update { force } => {
                 update::run_update(VERSION, force);
                 Ok(())
@@ -91,10 +109,6 @@ impl Action {
             }
             Action::ShowAvailableVersions => {
                 update::show_available_versions(VERSION);
-                Ok(())
-            }
-            Action::ShowConfig => {
-                Config::load().print_table();
                 Ok(())
             }
         }
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,6 +1,6 @@
 use std::collections::HashMap;
 
-use anaconda_otel_rs::signals::increment_counter;
+use anaconda_otel_rs::signals::{increment_counter, shutdown_telemetry};
 use clap::{Parser, Subcommand};
 use indoc::formatdoc;
 
@@ -10,9 +10,17 @@ use crate::config::{self, Config};
 use crate::update;
 
 pub fn execute() {
+    tracing_subscriber::fmt()
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .init();
+
     config::setup_telemetry();
 
-    if let Err(e) = parse().execute() {
+    let result = parse().execute();
+
+    shutdown_telemetry();
+
+    if let Err(e) = result {
         eprintln!("Error: {}", e);
         std::process::exit(1);
     }
@@ -57,16 +65,16 @@ impl Action {
         let name = self.match_action_name();
         let mut attrs = HashMap::new();
         attrs.insert("command".to_string(), name.into());
-        increment_counter("cli.command.invoked", 1, attrs.clone());
+        increment_counter("cli_command_invoked", 1, attrs.clone());
 
         let result = self.run();
 
         match &result {
             Ok(_) => {
-                increment_counter("cli.command.success", 1, attrs);
+                increment_counter("cli_command_success", 1, attrs);
             }
             Err(_) => {
-                increment_counter("cli.command.failure", 1, attrs);
+                increment_counter("cli_command_failure", 1, attrs);
             }
         }
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,7 +1,11 @@
+use std::collections::HashMap;
+
+use anaconda_otel_rs::signals::increment_counter;
 use clap::{Parser, Subcommand};
 use indoc::formatdoc;
 
 use crate::VERSION;
+use crate::update;
 
 /// Action to be performed, returned by parse()
 pub enum Action {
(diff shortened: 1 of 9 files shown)


src/cli.rs
```diff
@@ -14,6 +19,74 @@ pub enum Action {
...
+impl Action {
+    pub fn name(&self) -> &'static str {
+        match self {
+            Action::ShowHelp => "help",
+            Action::ShowSelfHelp => "self.help",
+            Action::ShowVersion => "version",
+            Action::ShowConfig => "config",
+            Action::Update { .. } => "self.update",
+            Action::CheckForUpdate => "self.update.check",
+            Action::ShowAvailableVersions => "self.update.list",
```
mattkram, line 59: This looks pretty good to me. I do wonder if there's a pattern to keep these names associated with the enum definitions, but that doesn't have to happen now. It's more of a long-term maintenance thing.

I do think having the name of this function be more specific to `telemetry_name` or something would be good. Does it have to be `pub`?

src/main.rs
```diff
@@ -1,18 +1,48 @@
...
+
 mod cli;
 mod config;
 mod update;
 
 pub const VERSION: &str = env!("PKG_VERSION");
 
+fn setup_telemetry() -> Result<(), Box<dyn std::error::Error>> {
+    let mut config = Configuration::new(
+        Some("https://metrics.auth.anacondaconnect.com/v1/metrics"),
```
mattkram: We should move this to `config.rs` to allow us to override for development too.

src/main.rs
```diff
@@ -1,18 +1,48 @@
...
+        None,
+    )?;
+
+    // Set auth token from environment variable
+    let api_key = env::var("OTEL_API_KEY").ok();
```
mattkram: Once we merge #50 we can load this via the keyring. Once we build out the full config system, we'll just provide with a common helper function.

src/main.rs
```diff
@@ -1,18 +1,48 @@
...
+    config.set_auth_token(api_key);
+
+    // Disable console exporter to use OTLP HTTP
+    config.set_console_exporter(false);
+
+    // Set a short export interval (1 second) so we don't have to wait 60s to see results
+    config.set_metrics_export_interval_ms(1000);
```
mattkram: Let's put this into `config.rs` too.

src/main.rs
```diff
@@ -1,18 +1,48 @@
...
-        cli::Action::ShowHelp => cli::print_main_help(),
-        cli::Action::ShowSelfHelp => cli::print_self_help(),
-        cli::Action::ShowVersion => println!("{}", VERSION),
-        cli::Action::ShowConfig => config::Config::load().print_table(),
-        cli::Action::Update { force } => update::run_update(VERSION, force),
-        cli::Action::CheckForUpdate => update::check_for_update(VERSION),
-        cli::Action::ShowAvailableVersions => update::show_available_versions(VERSION),
+    let _ = setup_telemetry();
+
+    if let Err(e) = cli::parse().execute() {
```
mattkram: I like this pattern. I'd also be cool with just implementing a `cli::execute()` that does the stacking of these two calls.

src/cli.rs
zfralish: I think it should probably be `action_name` or `match_action_name` or something as that is what the function is doing.

I'll see if anything breaks if i don't make it pub.

src/main.rs
zfralish: Yeah I was thinking to let you merge that first, then ill come and deal with conflicts and merge this one

src/config.rs
```diff
@@ -5,26 +5,62 @@
...
-//! | `ANA_SSL_VERIFY`     | `true`          | SSL certificate verification   |
-//! | `ANA_OPEN_BROWSER`   | `true`          | Auto-open browser during login |
-//! | `ANA_USE_HTTPS`      | `true`          | Use HTTPS (set false for HTTP) |
+//! | Variable                         | Default                    | Description                    |
+//! |--------------------------------- |----------------------------|--------------------------------|
+//! | `ANA_DOMAIN`                     | `anaconda.com`             | Authentication domain          |
+//! | `ANA_AUTH_CLIENT_ID`             | (Anaconda's ID)            | OAuth client ID                |
+//! | `ANA_SSL_VERIFY`                 | `true`                     | SSL certificate verification   |
+//! | `ANA_OPEN_BROWSER`               | `true`                     | Auto-open browser during login |
+//! | `ANA_METRICS_ENDPOINT`           | (Anaconda metrics URL)     | OpenTelemetry metrics endpoint |
```
mattkram: We should add a top-level `ANA_ENABLE_TELEMETRY`, defaulting to true. Now or in another.

src/update.rs
```diff
@@ -433,10 +433,8 @@ mod tests {
     #[test]
     fn test_get_asset_name_returns_platform_specific_name() {
         // This test verifies get_asset_name returns a valid result on supported platforms
-        let result = get_asset_name();
         // On CI/dev machines (macOS arm64, linux x86_64, windows x86_64), this should succeed
-        if result.is_ok() {
-            let name = result.unwrap();
+        if let Ok(name) = get_asset_name() {
```
mattkram, line 419: Nice

src/config.rs
```diff
@@ -5,26 +5,62 @@
...
+
+    Ok(())
+}
+
 // TODO(mattkram): Update default to anaconda.com before public release
 const DEFAULT_DOMAIN: &str = "stage.anaconda.com";
 const DEFAULT_CLIENT_ID: &str = "b4ad7f1d-c784-46b5-a9fe-106e50441f5a";
 const DEFAULT_SSL_VERIFY: bool = true;
 const DEFAULT_OPEN_BROWSER: bool = true;
+const DEFAULT_METRICS_ENDPOINT: &str = "https://metrics.auth.anacondaconnect.com/v1/metrics";
```
mattkram, line 68: We have to make sure that this aligns with the `DEFAULT_DOMAIN` above, because the tokens have to match. I think we should either keep it as staging for now, or derive it from `DEFAULT_DOMAIN` if it is predictable.

src/config.rs
```diff
@@ -42,6 +78,17 @@ pub struct Config {
...
+    pub metrics_endpoint: String,
+
+    /// Metrics export interval in milliseconds
+    pub metrics_export_interval_ms: i64,
+
+    /// Enable console metrics exporter
+    pub metrics_console_exporter: bool,
+
+    /// Skip internet connectivity check
+    pub metrics_skip_internet_check: bool,
```
mattkram, line 100: Nit
```suggestion
    pub metrics_skip_internet_check: bool,
    
```

src/cli.rs
```diff
@@ -1,7 +1,22 @@
...
+use crate::update;
+
+pub fn execute() {
+    config::setup_telemetry();
+
+    if let Err(e) = parse().execute() {
+        eprintln!("Error: {}", e);
+        std::process::exit(1);
+    }
+}
```
mattkram, line 27: I'm now debating again whether we want this in main or not. I'm leading towards yes, because that's where we should put general "startup" behavior.

I'm not dogmatic about it though. What do you think?

src/cli.rs
```diff
@@ -19,6 +34,87 @@ pub enum Action {
...
+        let name = self.match_action_name();
+        let mut attrs = HashMap::new();
+        attrs.insert("command".to_string(), name.into());
+        increment_counter("cli.command.invoked", 1, attrs.clone());
+
+        let result = self.run();
+
+        match &result {
+            Ok(_) => {
+                increment_counter("cli.command.success", 1, attrs);
```
mattkram: Does the otel library handle associating this event with the same command invocation as the previous one? I'm not familiar enough to know.

src/config.rs
```diff
@@ -5,26 +5,62 @@
...
+    let _ = try_setup_telemetry();
+}
+
+fn try_setup_telemetry() -> Result<(), Box<dyn std::error::Error>> {
+    let app_config = Config::load();
+
+    let mut otel_config = Configuration::new(Some(&app_config.metrics_endpoint), None)?;
+
+    // TODO: Add API key here
+    otel_config.set_auth_token(None);
```
mattkram: We should have enough to do this now.

src/cli.rs
```diff
@@ -1,7 +1,22 @@
...
 use clap::{Parser, Subcommand};
 use indoc::formatdoc;
 
 use crate::VERSION;
+use crate::auth;
+use crate::config::{self, Config};
+use crate::update;
+
+pub fn execute() {
+    config::setup_telemetry();
```
mattkram, line 17: This is probably where we should guard based on global opt-out flag.

src/update.rs
zfralish: Claude caught it not me haha

src/config.rs
zfralish: I am not sure the tokens have to match, the metrics endpoint is just ripping out the header and sending it to the auth service to see if it is valid. Right now, this is the only url configured for metrics in dev, we will have to do an infra pr to setup a staging one if needed.

src/cli.rs
zfralish: I think that makes sense, its an extra layer of abstraction for nothing really.

src/cli.rs
zfralish: Your query would handle that part, its groups and subgroups, so the main group would be `cli.command.success`, then under that we would have `whoami`, so in the query if you wanted to see how many time its successfully been called you would aggregate them and add all the int values into a sum.

I think 🤔

src/config.rs
zfralish: Donezo

src/cli.rs
zfralish: I put it inside the function, it will just fail silently in the background if the flag is false

src/config.rs
mattkram: I misinterpreted this as the prod URL. We are good for now :)

mattkram (approved):
Awesome! Code looks great, and I was able to validate it end-to-end and see telemetry flowing through!

:shipit:

PR #59, opened 2026-04-01
refac: Auto-generate config table from struct fields

zfralish:
## Summary
- Add `Serialize` derive to `Config` struct
- Use `serde_json::to_value` to iterate over struct fields automatically
- Remove manual row additions and unused `bool_to_str` helper

This eliminates the need to manually update `print_table` when adding new config options.

## Test plan
- [x] `cargo check` passes
- [x] `cargo test` passes (61 tests)
- [x] Verified `ana config` displays all fields correctly


mattkram (approved):
Nice. I like this. This also makes `ana config --json` pretty easy. I think that's going to be a pattern we'll want pretty broadly for different types of pretty or json output

PR #61, opened 2026-04-01
feat: Add ability to bootstrap a standalone anaconda CLI environment

mattkram:
## Summary

This is a POC, heavily inspired by the approach used in https://github.com/jezdez/conda-express, to bootstrap a standalone environment containing the Anaconda CLI (implemented in Python). In order to do this, we statically embed a cross-platform lockfile and install using `rattler`. When the user runs the bootstrap command, they get our CLI packages, plus a Python runtime, isolated from all other system conda/python environments.

At this point, we are doing this for a few reasons:

* To better understand the underlying technology to understand edges and learn
* To prototype a method for rapidly shipping existing CLI functionality, all bootstrapped from `ana`, without having to do a full rewrite
* As a foundation for a **possible** general mechanism for tool installation

### How it works

* Adds a new subcommand `ana bootstrap`, which invokes the bootstrapping process
* There is a step included as well to do this in the `install.sh` script, after the `ana` binary is installed
* We store a rattler v6 lockfile, which is statically compiled into `ana` at build time. This could be extended in the future to fetch a package spec from a service for things like custom tools, etc.
* We use `rattler` to install a new environment inside `~/.ana/tools/anaconda-cli`
* We symlink the binary into `~/.ana/bin`. This is configurable per-tool, in cases where multiple entrypoints might need to be exposed (inside `lockfiles.rs`)
* The `~/.ana/bin` directory is also put onto `PATH` inside the `install.sh` script
* Adds a simple `ana org ...` wrapper, which runs `anaconda org ...` in a subprocess. This is just to demonstrate what a subprocess-level wrapper might look like

### Some design details

* The lockfile is configured specifically to use https://repo.anaconda.com (as opposed to conda-forge). We add `anaconda-cloud` channel on anaconda.org just because some of the packages are not in main yet
* One reason to pursue this approach is we can configure where the packages for the tool environments come from. In the future, with some backend services, we can potentially make this dynamic based on user/customer preferences
* We don't do any channel configuration yet

The binary size is now 15MB, compared to ~10MB of the current `main` branch build. Primary size increase is the one time addition of the rattler crates. Preliminary testing shows addition of more lockfiles has negligible effect on the binary size.

### Benchmarking

Preliminary benchmarking shows the bootstrapping process completing in <2s, with a cleared rattler cache: 
image

### Known issues

* On macOS, the first execution of `anaconda` is very slow. This is due to a combination of byte-compiling the Python code, but also most likely due to the Gatekeeper security feature of macOS, because it considers the entire folder as something that needs to be scanned.
* This does not (yet) do any shell activation, to the same extent that `pixi run` and `pixi shell` do. For `anaconda` CLI packages this should be fine, but this needs to be solved more generically if we decide to extend this feature more generally.

Jira: [CLI-377](https://anaconda.atlassian.net/browse/CLI-377)

[CLI-377]: https://anaconda.atlassian.net/browse/CLI-377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -398,6 +398,15 @@ run_bootstrap() {
 }
 
 update_shell_profile() {
+    local _dir="$1"
+    local _ana_bin_dir="$HOME/.ana/bin"
+
+    # Add both the install directory and ~/.ana/bin (for tool symlinks)
+    add_to_path "$_dir"
+    add_to_path "$_ana_bin_dir"
+}
+
+add_to_path() {
     local _dir="$1" _line
 
     # Already in $PATH
@@ -424,7 +433,6 @@ update_shell_profile() {
             return 0
             ;;
     esac
-    printf "\n"
 }
 
 append_line_if_missing() {
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -47,6 +47,7 @@ Options:
   -v, --version VERSION    Version to install (default: ${DEFAULT_VERSION})
       --no-verify-checksum Disable checksum validation after download (default: false)
       --no-path-update     Skip shell profile modification
+      --no-bootstrap       Skip running 'ana bootstrap' after installation
   -t, --token TOKEN        GitHub token for private repo access
   -f, --force              Overwrite existing installation without prompting
   -h, --help               Show this help message
@@ -56,6 +57,7 @@ Environment variables:
   ANA_VERSION              Same as --version
   ANA_VERIFY_CHECKSUM      Set to "false" to skip checksum verification
   ANA_NO_PATH_UPDATE       Set to non-empty to skip PATH update
+  ANA_BOOTSTRAP            Set to "false" to skip bootstrap
   ANA_FORCE_INSTALL        Set to non-empty to overwrite without prompting
   GITHUB_TOKEN             Same as --token
 
@@ -99,6 +101,10 @@ parse_args() {
                 ;;
             --no-path-update)
                 ANA_NO_PATH_UPDATE="1"
+                shift
+                ;;
+            --no-bootstrap)
+                ANA_BOOTSTRAP="false"
                 shift
                 ;;
             -f|--force)
@@ -175,6 +181,8 @@ main() {
     if [ -z "${ANA_NO_PATH_UPDATE:-}" ]; then
         update_shell_profile "$_install_dir"
     fi
+
+    run_bootstrap "$_install_dir"
 
     printf "🎉 Done! Run '\033[1;36mana --help\033[0m' to get started.\n"
 }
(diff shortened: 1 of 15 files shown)


scripts/install.sh
```diff
@@ -365,7 +373,40 @@ install_binary() {
...
+            ;;
+        true|1) ;;
+        *)
+            err "Invalid ANA_BOOTSTRAP value '%s'. Must be 'true', 'false', '1', or '0'." "$_bootstrap"
+            ;;
+    esac
+
+    info "Running ana bootstrap..."
+    if "$_ana_bin" bootstrap; then
+        info "Bootstrap completed successfully"
```
zfralish, line 394: Noticed other tools output the path that is attempted to be added to PATH, that way people who use other shells can manually do it easily.

src/tools/install.rs
```diff
@@ -0,0 +1,209 @@
...
+    package_cache::PackageCache,
+};
+use rattler_conda_types::{Platform, PrefixRecord};
+use rattler_lock::LockFile;
+
+use crate::paths;
+use super::lockfiles;
+
+/// Global progress bar for installation feedback.
+static MULTI_PROGRESS: std::sync::LazyLock<MultiProgress> = std::sync::LazyLock::new(|| {
```
zfralish, line 19: This might want to go in a `components` or `ui` file unless it is install specific.

src/tools/install.rs
```diff
@@ -0,0 +1,209 @@
...
+    let package_cache = PackageCache::new(cache_dir.join(rattler_cache::PACKAGE_CACHE_DIR));
+
+    // Run installation
+    let start = Instant::now();
+    let result = Installer::new()
+        .with_download_client(client)
+        .with_package_cache(package_cache)
+        .with_target_platform(platform)
+        .with_installed_packages(installed)
+        .with_execute_link_scripts(true)
```
zfralish, line 95: Should this always be true, would we like to to be dynamic like the others?

src/anaconda_cli.rs
```diff
@@ -0,0 +1,48 @@
...
+
+pub fn run_bootstrap() -> Result<(), String> {
+    let anaconda_bin = paths::bin_dir().join("anaconda");
+
+    if anaconda_bin.exists() {
+        eprintln!("anaconda-cli is already installed");
+        return Ok(());
+    }
+
+    eprintln!("Installing anaconda-cli...");
```
zfralish, line 14: Might be nice to reuse that progress bar from the other file

zfralish (approved):
I don't see anything glaring, but bash is also not my strong suite, might want to find a review from a bash wiz as well.

Also, is there a way we could add a bunch of debug logs that are supressed by default?

src/tools/install.rs
mattkram: This one is specific to the interface of rattler, but it's a good idea if we find we need it more later.

src/anaconda_cli.rs
mattkram: Opened in #68

scripts/install.sh
mattkram: This is handled to some extent below in `add_to_path`, but I like the idea of reviewing the output messages again.

mattkram: Opened #69

src/tools/install.rs
mattkram: Added a TODO to remind us to review this question later

PR #70, opened 2026-04-03
feat: Add system telemetry attributes

zfralish:
## Summary
- Add `os`, `arch`, and `version` attributes to all telemetry metrics
- Uses `std::env::consts` for OS and architecture detection (no new external dependencies for system info)
- Added `opentelemetry` as direct dependency for the `Value` type

## Test plan
- [x] `cargo build` compiles successfully
- [x] `cargo test` - all 61 tests pass
- [ ] Verify telemetry metrics include new attributes in backend

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ thiserror = "2"
 tokio = { version = "1.45", features = ["rt-multi-thread", "macros"] }
 webbrowser = "1"
 tracing-subscriber = { version = "0.3.23", features = ["env-filter"] }
+opentelemetry = "0.31.0"
 
 [dev-dependencies]
 temp-env = "0.3"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -23,6 +23,7 @@ anaconda-otel-rs = { git = "https://github.com/anaconda/anaconda-otel-rs" }
 thiserror = "2"
 webbrowser = "1"
 tracing-subscriber = { version = "0.3.23", features = ["env-filter"] }
+opentelemetry = "0.31.0"
```
mattkram, line 33: Should this live in the `anaconda-otel-rs` (this PR) instead? I'm okay either way, but feels pretty "core".

mattkram (approved):

Cargo.toml
zfralish: Im not sure this is even needed, let me try removing it.

mattkram: You need it because of the direct `use opentelemetry::Value`, but the question is more, should this be in the base otel wrapper. For now it's okay here, so I'm gonna merge

mattkram (approved):

PR #80, opened 2026-04-03
feat: Add lockfile updates to the pre-commit job

mcg1969:
Here is a PR that I don't think will require as much thoughtful conversation :-) I noticed the building up renovate PRs, and in my experience, one problem with them is that they don't play nicely with pre-commit hooks, and they often fail to propagate package version updates to lockfiles properly. So this PR does two things:

- It adds Cargo.lock and pixi.lock regen to the same task that runs the pre-commit hooks
- Instructs renovate not to care if [devops+anaconda-bot@anaconda.com](mailto:devops+anaconda-bot@anaconda.com) adds a commit to their open PRs.

Hopefully with this fix in place it will be easier for y'all to accommodate the flood of renovate requests. Of course, in those cases where a renovate change breaks legitimate tests... well, that remains a manual issue.

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -33,6 +33,10 @@ jobs:
         with:
           token: ${{ secrets.ANACONDA_BOT_TOKEN }}
 
+      - name: Configure git for private repos
+        run: |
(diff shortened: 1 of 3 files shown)


mattkram (approved):
Fantastic, thanks!

PR #82, opened 2026-04-04
ci: Add Cargo caching and simplify conda packaging

mcg1969:
## Summary

- **Cargo caching**: Cache `~/.cargo/registry`, `~/.cargo/git`, and `target/` keyed on `Cargo.lock` hash, with OS-specific restore keys
- **Simplified conda packaging**: Pre-built binary packaging instead of compiling from source in rattler-build. Removes rust/pkg-config/openssl build deps from the recipe. Build scripts verify the binary version matches `PKG_VERSION` before packaging
- **Combined CI jobs**: Merge the separate build-and-test + build-conda jobs into a single `build` job per OS, reducing overhead and enabling cache reuse
- **Concurrency control**: Cancel in-progress CI runs when a new push arrives to the same PR
- **Makefile**: Flesh out Make targets to mirror pixi tasks
- **Rust toolchain**: Explicit `dtolnay/rust-toolchain@stable` setup for consistent toolchain management

## CI Timing Comparison

All timings are wall-clock per-OS job duration. Main branch runs two parallel jobs per OS (Build & Test + Build Conda); the wall-clock time is the slower of the two.

| OS | Main (2 jobs) | This PR, cold cache | This PR, warm cache |
|----|---------------|---------------------|---------------------|
| macOS | 5m 03s | 5m 19s | **1m 28s** |
| Ubuntu | 8m 45s | 7m 07s | **1m 46s** |
| Windows | 17m 14s | 17m 37s | **3m 28s** |

Cold cache performance is comparable to main (slightly faster on Ubuntu due to combined jobs, slightly slower on Windows due to cache save overhead). **Warm cache cuts CI time by 70-80%.**

## Test plan

- [x] CI passes on all three platforms (ubuntu, macos, windows)
- [x] Cold cache run succeeds (caches deleted, full rebuild)
- [x] Warm cache run succeeds (cache hit, incremental rebuild)
- [x] Conda package builds and uploads on all platforms
- [x] Binary uploads on all platforms
- [x] Pre-commit passes


conda.recipe/recipe.yaml
```diff
@@ -1,31 +1,13 @@
...
-      then:
-        - openssl
-
-tests:
-  - script:
-      - if: unix
-        then: run_test.sh
-      - if: win
-        then: run_test.bat
-
```
mattkram, line 28: Do we lose any robustness or binary compatibility guarantees by building outside of the recipe? That was one of the reasons I initially did it this way, based on an assumption that we may have some benefit of using a standard conda-based compiler stack. But admittedly I don't have the level of depth here required to fully be able to answer my own question.

In any case, since this is the development recipe it is probably sufficient to do as you propose, simply for the performance benefits of the shared cache. And if we find at any point we have any odd compatibility issues we can re-visit.

scripts/get_version.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+# Derive the package version from git tags and write it to VERSION
+# and conda.recipe/variants.yaml.
+#
+# Version format:
+#   Exact tag match (v0.1.0-0-ghash):        0.1.0
+#   Commits past a tag (v0.1.0-5-gabcdef1):  0.1.0+5.gabcdef1
```
mattkram: This is pretty similar to our current versioning, but it is not exactly the same. Specifically, in this example the `+5` should be `.dev5`. I'd prefer if we can find a way to at least match that convention until we have a more stable way to store our intermediate dev builds.

I'm also anticipating we will manually cut releases with tags liked `v0.1.0-rc.0`, as we get closer to "real" releases. `setuptools-scm` handles this well for us already, and I'd like to make sure we have at least a path towards achieving that.

If we want to just update to handle the `.devX` convention for now, that would be good with me.

Also, FWIW, I do like the `+g<hash>` part as well, but I had removed it just so our automated releases don't get it. Maybe there'd be a way to have those locally, but disable that part with an environment variable or flag to this script in CI (`release.yaml`)?

build.rs
```diff
@@ -1,12 +1,21 @@
...
-    println!("cargo:rerun-if-env-changed=PKG_VERSION");
+    // Cargo will re-run this build script whenever the VERSION file changes.
+    println!("cargo:rerun-if-changed=VERSION");
 
-    // Version is passed via PKG_VERSION environment variable.
-    // This should be set by CI or locally via: PKG_VERSION=$(python scripts/get_version.py)
-    // Falls back to CARGO_PKG_VERSION if not set.
-    let version =
-        std::env::var("PKG_VERSION").unwrap_or_else(|_| env!("CARGO_PKG_VERSION").to_string());
+    let version = std::fs::read_to_string("VERSION")
```
mattkram: It would be cool if `build.rs` could run the script (if we are on a POSIX system that supports it) in case users run `cargo build` manually themselves, locally.

mattkram (commented):
Overall, this looks really great, and I really like the direction you are going (and the performance benefits).

Two primary areas that I'd like to further explore:
* Does building outside the conda recipe lose us anything in terms of compatibility? Probably not, but wanted to raise the question for discussion.
* Is it possible to update the version generation script to better match the existing behavior (and future needs around `-alpha`, `-rc` versions?

conda.recipe/recipe.yaml
mcg1969: This is a good question. The _generic_ answer is that it depends. But for a Rust-built binary, that does not depend on conda-installed library dependencies, the answer is no. Here is the current dependency list on the Mac:
```
$ otool -L target/release/ana
target/release/ana:
	/System/Library/Frameworks/CoreServices.framework/Versions/A/CoreServices (compatibility version 1.0.0, current version 1226.0.0)
	/System/Library/Frameworks/IOKit.framework/Versions/A/IOKit (compatibility version 1.0.0, current version 275.0.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 4201.0.0)
	/System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 61901.60.40)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1356.0.0)
```
For me, the more important question is: _are we fully testing the thing we are shipping_? If you don't run the same tests inside the conda build that you're running outside of it, then you can't be truly sure that your package is fully tested. So for me, this is superior because we have fully tested this binary prior to putting it inside of the conda package. It's the same binary.

scripts/get_version.sh
mcg1969: I am happy to update the versioning to however you wish to work it. However, it's important to note a couple of things:

- that we can't use dashes in the conda version number.
- the `.devN` version tag behaves different than you maty think. According to conda, `1.0.0.dev5` actually comes _before_ `1.0.0`. On the other hand, `1.0.0+5` comes _after_ `1.0.0`. So if you use `devN` the way you're asking, you're getting the builds in the wrong order. Someone who wants the latest development can't get it with a `conda update`.
- as for the `+g<hash>` part, that's something I use all the time on intermediate conda builds, because the commit count does not guarantee the precise build alone. For tagged releases, that hash is dropped.

build.rs
mcg1969: I'll investigate this!

scripts/get_version.sh
mcg1969: To be clear I didn't mean to step on deliberate versioning choices on your part, sorry!

mattkram: No worries, at all. That is indeed how I expected them to work. If you look at our releases page, you can see the current automated releases coming off main: https://github.com/anaconda/ana-cli/releases.

Indeed, after a manually-created `v0.0.5`, they are starting at `0.0.6.dev1`, etc.

The primary decision to use `setuptools-scm` was because it handles the annoying complexities for us in a standard way. It has built-in handling as well for replacement of dashes to form well-formed and sortable SemVer strings. We do this in the python-based packages as well, which is where the choice originated (see e.g. https://anaconda.org/channels/anaconda-cloud/packages/anaconda-auth/files).

image

The other wrinkle in this is that the recommended tagging formats by GitHub releases, which we generally have used (i.e. begin with `v`, and use format like `v0.1.0-rc.0`), are not directly supported by pip or conda. But the rules in `setuptools-scm` handle this for us.

I wonder if there's a way we can still maintain the caching support without replacing all of the regex logic?

mcg1969: Ah, OK, so we bump the patch release before applying `.dev`; that makes sense

mcg1969: I'll bring back in setuptools-scm, that makes a lot of sense!

build.rs
mcg1969: For now I've punted on this in favor of a full reversion of the get_version.py logic. We can address that separately

conda.recipe/recipe.yaml
mattkram: I'm good with this. Thanks for helping me understand a bit more :)

mattkram (approved):
This looks great, thanks for doing it!

PR #85, opened 2026-04-06
feat: Add verbosity flags for tracing subscriber

zfralish:
## Summary
- Add `-v`/`--verbose` flag with incremental verbosity levels for logging
- Levels: `-v` (error), `-vv` (warn), `-vvv` (info), `-vvvv` (debug), `-vvvvv` (trace with all deps)
- Flag is global so it works with any subcommand (e.g., `ana -vvv login`)
- `RUST_LOG` env var still takes precedence when set

## Test plan
- [ ] Verify `ana --help` shows the new `-v, --verbose` option
- [ ] Test `-v` through `-vvvvv` produce appropriate log output
- [ ] Confirm `RUST_LOG=debug ana config` still overrides verbosity flags

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -13,6 +13,44 @@ use crate::auth;
 use crate::config::{self, Config};
 use crate::update;
 
+/// Log level for tracing output.
+#[derive(Debug, Clone, Copy, Default)]
+pub enum LogLevel {
+    #[default]
+    Off,
+    Error,
+    Warn,
+    Info,
+    Debug,
+    Trace,
+}
+
+impl From<u8> for LogLevel {
+    fn from(count: u8) -> Self {
+        match count {
+            0 => Self::Off,
+            1 => Self::Error,
+            2 => Self::Warn,
+            3 => Self::Info,
+            4 => Self::Debug,
+            _ => Self::Trace,
+        }
+    }
+}
+
+impl LogLevel {
+    fn as_filter_str(&self) -> &'static str {
+        match self {
+            Self::Off => "off",
+            Self::Error => "ana=error,anaconda_otel_rs=off,opentelemetry=off,reqwest=off",
+            Self::Warn => "ana=warn,anaconda_otel_rs=off,opentelemetry=off,reqwest=off",
+            Self::Info => "ana=info,anaconda_otel_rs=off,opentelemetry=off,reqwest=off",
+            Self::Debug => "ana=debug,anaconda_otel_rs=off,opentelemetry=off,reqwest=off",
+            Self::Trace => "ana=trace,anaconda_otel_rs=off,opentelemetry=off,reqwest=off",
+        }
+    }
+}
+
 /// Build base telemetry attributes with system information.
 fn system_attrs() -> HashMap<String, Value> {
     let mut attrs = HashMap::new();
@@ -23,9 +61,9 @@ fn system_attrs() -> HashMap<String, Value> {
 }
 
 pub async fn execute() {
-    let (action, verbosity) = parse();
+    let (action, level) = parse();
 
-    let filter = build_tracing_filter(verbosity);
+    let filter = build_tracing_filter(level);
     tracing_subscriber::fmt().with_env_filter(filter).init();
 
     config::setup_telemetry();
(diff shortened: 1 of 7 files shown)


src/cli.rs
```diff
@@ -23,24 +23,43 @@ fn system_attrs() -> HashMap<String, Value> {
...
     if let Err(e) = result {
+        tracing::error!("Command failed: {}", e);
         eprintln!("Error: {}", e);
         std::process::exit(1);
     }
 }
 
+/// Build tracing filter based on verbosity level.
+/// Respects RUST_LOG env var if set, otherwise uses verbosity flags.
+fn build_tracing_filter(verbosity: u8) -> tracing_subscriber::EnvFilter {
```
mattkram: Is there any way, or use in defining an enum mapping integers to log levels? This is probably fine for now, but if we need to restrict ourselves to integers we should encapsulate that in one location and do our best not to have magic ints floating around.

src/cli.rs
```diff
@@ -148,48 +167,52 @@ impl Action {
     }
 }
 
-/// Parse CLI arguments and return the action to perform.
+/// Parse CLI arguments and return the action to perform along with verbosity level.
 /// Exits the process on unrecoverable errors (unknown commands, etc.)
-pub fn parse() -> Action {
+pub fn parse() -> (Action, u8) {
```
mattkram: Maybe this is the function where we should convert `u8` -> `LogLevel::Info` or something?

src/update.rs
```diff
@@ -12,7 +12,10 @@ const GITHUB_REPO: &str = "anaconda/ana-cli";
 fn github_client() -> Result<reqwest::Client, Error> {
     let token = match env::var("GITHUB_TOKEN") {
         Ok(token) if !token.is_empty() => token,
-        _ => return Err(Error::MissingToken),
+        _ => {
+            tracing::error!("GITHUB_TOKEN not set or empty");
+            return Err(Error::MissingToken);
```
mattkram, line 17: Maybe a future thing, but I wonder if there are crates or other patterns that can have a string representation defined to pass into `tracing::error!` automatically.

mattkram (approved):
Looks good to me. Just one comment about the log level integer, which doesn't have to block but I think would be a good thing to add.

src/cli.rs
zfralish: https://github.com/anaconda/ana-cli/pull/85/changes#diff-b2812f19576dd53d0c35b107a322f58a00fce6977f4b5976e1961853982af3ccR18

This does that, in most places it is this enum type, or just an int.

The only `magic` place is where that conversion `From<u8>` hapens.

src/update.rs
zfralish: Yeah would be nice to just have the error it self parsed into a string as well

src/cli.rs
zfralish: This is how it currently works, where are you suggesting a change should be made?

```
  1. Clap parses -v flags (cli.rs:340-342)
  #[arg(short = 'v', long = "verbose", action = clap::ArgAction::Count, global = true)]
  verbose: u8,
  1. ArgAction::Count increments for each -v. So -vvv → 3u8.
  2. parse() converts to LogLevel (cli.rs:201-204)
  pub fn parse() -> (Action, LogLevel) {
      match Cli::try_parse() {
          Ok(cli) => {
              let level: LogLevel = cli.verbose.into();
  2. Calls the From<u8> impl.
  3. From<u8> maps count to variant (cli.rs:28-39)
  impl From<u8> for LogLevel {
      fn from(count: u8) -> Self {
          match count {
              0 => Self::Off,
              1 => Self::Error,
              2 => Self::Warn,
              3 => Self::Info,
              4 => Self::Debug,
              _ => Self::Trace,
          }
      }
  }
  4. execute() receives the level (cli.rs:63-67)
  pub async fn execute() {
      let (action, level) = parse();
      let filter = build_tracing_filter(level);
      tracing_subscriber::fmt().with_env_filter(filter).init();
  5. build_tracing_filter() checks env first, then uses level (cli.rs:84-90)
  fn build_tracing_filter(level: LogLevel) -> tracing_subscriber::EnvFilter {
      if let Ok(filter) = tracing_subscriber::EnvFilter::try_from_default_env() {
          return filter;
      }
      tracing_subscriber::EnvFilter::new(level.as_filter_str())
  }
  6. as_filter_str() returns the directive (cli.rs:41-52)
  fn as_filter_str(&self) -> &'static str {
      match self {
          Self::Off => "off",
          Self::Error => "ana=error,anaconda_otel_rs=off,opentelemetry=off,reqwest=off",
          // ...
      }
  } 
```

zfralish: Oh I see  you made the change

zfralish: Joys of vibe coding haha

PR #86, opened 2026-04-06
feat: Add request/response logging middleware for HTTP clients

zfralish:
## Summary
- Add logging middleware using `reqwest-middleware` that logs HTTP requests and responses
- Logs outgoing requests (method + URL) at debug level
- Logs successful responses (method + URL + status) at debug level  
- Logs failed requests at error level
- Update all auth HTTP clients to use the middleware

## Dependencies
This PR is based on #85 (verbosity flags) and should be merged after it.

## Test plan
- [ ] Run `ana -vvvv login` and verify request/response logs appear
- [ ] Run `ana -vvvv whoami` and verify request/response logs appear
- [ ] Verify normal operation (without -v flags) shows no extra output


src/auth/actions.rs
```diff
@@ -307,15 +306,20 @@ pub fn show_api_key() -> Result<(), AuthError> {
...
-        println!("Not logged in to {}", client.domain());
+    if !api.is_authenticated() {
+        println!("Not logged in to {}", api.domain());
         println!("Run `ana login` to authenticate.");
         return Ok(());
     }
 
-    let response = client.get("/api/account").await?;
+    let response = api
+        .client
```
mattkram: Personally, I like keeping the local variable here named `client` (of type `ApiClient`). I think it feels more natural, and that we should then expose `client.get()` wrappers around a private instance of the `request::Client` instance.

src/auth/actions.rs
```diff
@@ -25,9 +26,10 @@ fn print_qr(qr: &str) {
 
 /// HTTP client with configuration and optional authentication.
 pub struct ApiClient {
-    client: reqwest::Client,
-    config: Config,
+    /// The underlying HTTP client with base URL and logging middleware.
+    pub client: Client,
```
mattkram: I like the storing of the base URL, and adding the logging middleware. Personally, I think we should still keep the underlying Client instance private and expose its methods via wrappers. This allows consumers to have a simple high-level interaction with the `ApiClient` type, while allowing it to wrap the complexity of URL handling, error handling, and logging.

mattkram (commented):
I like the idea of centralizing the core middleware etc. in a single type. After reviewing (sort of backwards), I'm wondering if we have too many layers here? I'm not fundamentally opposed, if we need that. But I would approach this with the assumption that the primary interface that consumers will need is the `ApiClient` type. And everything else should be more or less handled internally.

src/auth/actions.rs
zfralish: This should be resolved in the most recent commit.

src/auth/actions.rs
zfralish: This should be fixed in most recent commit

src/auth/api_keys.rs
```diff
@@ -35,12 +36,17 @@ pub async fn create_api_key(
...
 
+    let body = serde_json::to_string(&payload)
+        .map_err(|e| AuthError::Network(format!("Failed to serialize request: {}", e)))?;
+
     let response = client
         .post(&url)
         .bearer_auth(access_token)
-        .json(&payload)
+        .header(header::CONTENT_TYPE, "application/json")
+        .body(body)
```
mattkram: Is there no `.json()` method on this type anymore? It would be helpful if there was.

mattkram (approved):
Looks great. One small comment, but not critical

src/auth/api_keys.rs
mattkram: Turns out there is, but it requires a `json` feature for the `reqwest-middleware` crate. I mad the change.

mcg1969 (commented):
Nice foundation for centralized HTTP handling. Two things I'd like to see addressed before merging:

### 1. Unmigrated HTTP clients

Two files still create raw `reqwest::Client` instances, bypassing the logging middleware:

- **`src/update.rs:12-31`** — `github_client()` builds a `reqwest::Client` directly with user-agent and auth headers. This could use `build_client()` and return `ClientWithMiddleware` instead. The two callsites (lines 117, 172) use `.get().send().await` which works the same on `ClientWithMiddleware`.
  - **Complexity: low** — change `reqwest::Client::builder()...build()` to `build_client(reqwest::Client::builder()...)`, update return type and error mapping. ~10 lines changed.

- **`src/tools/install.rs:179-188`** — `make_download_client()` already wraps in `reqwest_middleware::ClientBuilder` but doesn't add `LoggingMiddleware`. It also has a TODO for `AuthenticationMiddleware`. This could use `build_client()` directly, though note it needs `.no_gzip()` on the inner builder.
  - **Complexity: trivial** — replace the manual `reqwest_middleware::ClientBuilder::new(client).build()` with `build_client(reqwest::Client::builder().user_agent(...).no_gzip())`. ~3 lines changed.

### 2. User-agent should be auto-applied

Every callsite currently has to remember to add `.user_agent(crate::ua::user_agent())` to the builder before passing it to `Client::new()` or `build_client()`. Since the whole point of centralizing is consistency, consider having `build_client()` / `Client::new()` apply the user-agent automatically. This makes it impossible to forget, and callers that need a custom user-agent (I don't think we have any) could override it via `default_headers`.

  - **Complexity: low** — add `.user_agent(crate::ua::user_agent())` inside `build_client()` before `.build()`, then remove it from the 4 callsites. ~8 lines changed total.

PR #90, opened 2026-04-06
feat: Add static user-agent string with platform identification

mcg1969:
## Motivation

Without a user-agent string, Cloudflare's heuristics sometimes classify our HTTP requests as malicious and block them. Adding a descriptive UA string prevents this.

## Summary

- Add a static user-agent string: `ana/{version} {platform} rattler/{version}`
- Platform detection aligns with conda's user-agent conventions (kernel, OS distribution, glibc)
- All detection uses direct syscalls or file reads — no subprocesses
- Rattler version extracted from `Cargo.lock` at compile time via `build.rs`
- String computed once per process via `LazyLock` and exposed as `ua::user_agent()`

## CI output

```
Linux:   ana/0.0.6.dev15 Linux/6.17.0-1008-azure ubuntu/24.04 glibc/2.39 rattler/0.40.3
macOS:   ana/0.0.6.dev15 Darwin/24.6.0 OSX/15.7.4 rattler/0.40.3
Windows: ana/0.0.6.dev15 Windows/10.0.26100 rattler/0.40.3
```

## Test plan

- [x] `cargo test` — all tests pass on Linux, macOS, and Windows
- [x] CI green on all three platforms

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,9 @@ tracing = "0.1"
 tracing-subscriber = { version = "0.3.23", features = ["env-filter"] }
 opentelemetry = "0.31.0"
 
+[build-dependencies]
+cargo-lock = "11"
+
 [dev-dependencies]
 temp-env = "0.3"
 tempfile = "3"
--- a/build.rs
+++ b/build.rs
@@ -18,4 +18,24 @@ fn main() {
     // Expose build target info for Sentry tags
     let target = std::env::var("TARGET").unwrap_or_else(|_| "unknown".to_string());
     println!("cargo:rustc-env=BUILD_TARGET={}", target);
+
+    // Extract rattler version from Cargo.lock for the user-agent string
+    println!("cargo:rerun-if-changed=Cargo.lock");
+    let rattler_version =
+        extract_dep_version("Cargo.lock", "rattler").unwrap_or_else(|| "unknown".to_string());
+    println!("cargo:rustc-env=RATTLER_VERSION={}", rattler_version);
+}
+
+/// Extract a dependency's resolved version from Cargo.lock.
+fn extract_dep_version(lock_path: &str, dep_name: &str) -> Option<String> {
+    let content = std::fs::read_to_string(lock_path).ok()?;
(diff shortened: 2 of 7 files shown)


pixi.toml
```diff
@@ -33,7 +33,7 @@ cmd = "python scripts/with_version.py cargo build --release"
 description = "Build the standalone Rust binary in release mode"
 
 [tasks.test]
-cmd = "python scripts/with_version.py cargo test"
+cmd = "python scripts/with_version.py cargo test -- --show-output"
```
mcg1969, line 36: This is so we get the output of the tests printed, one of which includes the user agent, so we can manually inspect it (or, in my case, ask Claude to extract it from the logs and put it in the PR text above.)

build.rs
```diff
@@ -18,4 +18,24 @@ fn main() {
...
+    let content = std::fs::read_to_string(lock_path).ok()?;
+    let mut found_name = false;
+    for line in content.lines() {
+        if line.starts_with("name = ") && line.contains(&format!("\"{}\"", dep_name)) {
+            found_name = true;
+        } else if found_name && line.starts_with("version = ") {
+            return line.split('"').nth(1).map(|s| s.to_string());
+        }
+    }
+    None
```
mattkram: I suspect that a more robust method for this would be to parse the toml format. And then to just iterate through the array of packages, and index by name and version.

Cargo.toml
```diff
@@ -41,6 +41,9 @@ tracing = "0.1"
 tracing-subscriber = { version = "0.3.23", features = ["env-filter"] }
 opentelemetry = "0.31.0"
 
+[build-dependencies]
+cargo-lock = "11"
```
mcg1969, line 45: nice! i'm not familiar with this crate (nor really too many others)

mattkram (approved):
Simplified `build.rs` by using the `cargo-lock` crate, but other than that looks great!

build.rs
mattkram: Replaced this with https://crates.io/crates/cargo-lock

PR #94, opened 2026-04-07
fix: Add user-agent to all HTTP clients and enable bootstrap CI tests

mcg1969:
## Summary

Several HTTP clients were missing user-agent headers. The package download client (`make_download_client()`) had no user-agent at all, causing `403 Forbidden` from `repo.anaconda.com` and breaking `ana bootstrap` in CI. The update client had a hardcoded `"ana-cli"` string instead of the standard UA. The auth clients (ApiClient and login flow) had no user-agent.

This PR adds `ua::user_agent()` to all HTTP clients and removes the `requires_anaconda_repo` CI skip marker so bootstrap tests actually run.

#90 added the `ua` module but only wired it into the static user-agent header — it missed these client construction sites.

## What changes

- `src/tools/install.rs`: Add `.user_agent(crate::ua::user_agent())` to rattler download client
- `src/update.rs`: Replace hardcoded `"ana-cli"` with `ua::user_agent()`
- `src/auth/actions.rs`: Add user-agent to ApiClient and login device-auth flow
- `tests/test_cli.py`: Remove `requires_anaconda_repo` skip decorator from 4 bootstrap tests

The following tests now run in CI:
- `test_bootstrap_installs_anaconda_cli`
- `test_bootstrap_creates_symlinked_binary`
- `test_bootstrap_already_installed`
- `test_bootstrap_anaconda_binary_runs`

## Test plan

- [x] CI passes on all platforms (macOS, Linux, Windows)
- [x] Bootstrap tests no longer skipped

### DIFF
--- a/src/auth/actions.rs
+++ b/src/auth/actions.rs
@@ -45,6 +45,7 @@ impl ApiClient {
         }
 
         let client = reqwest::Client::builder()
+            .user_agent(crate::ua::user_agent())
             .timeout(REQUEST_TIMEOUT)
             .default_headers(default_headers)
             .build()?;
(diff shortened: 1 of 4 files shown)


mattkram (approved):
Good catch. I missed the test bypasses 😬

PR #100, opened 2026-04-08
fix: Use real UTC timestamp in SBOM and sanitize local paths

mcg1969:
## Summary

Two SBOM processing fixes:

- **Real UTC timestamp**: Replace the fixed `1970-01-01T00:00:00Z` epoch with the actual generation time. The original `SOURCE_DATE_EPOCH=0` was used to produce reproducible per-target SBOMs, but this leaked into the final merged SBOM as a misleading timestamp. The `material_content()` comparison already excludes metadata, so timestamp-only changes do not trigger unnecessary SBOM rewrites — the reproducibility concern is fully addressed at the comparison layer.

- **Complete path sanitization**: The previous sanitization only handled `metadata.component.bom-ref` and `purl`, missing the `dependencies[].ref` field. Replaced with a global regex substitution that rewrites every occurrence of the developer's local workspace path (e.g. `path+file:///Users/foo/Repos/ana-cli#...`) to a generic `path+file:///tmp/ana-cli#...`.

## Test plan

- [ ] Pre-commit SBOM freshness hook passes
- [ ] `pixi run sbom` reports "No material changes" on re-run (timestamp-only change correctly skipped)
- [ ] `grep -c '/Users/' SBOM.json` returns 0

### DIFF
--- a/SBOM.json
+++ b/SBOM.json
@@ -2,8 +2,9 @@
   "bomFormat": "CycloneDX",
   "specVersion": "1.4",
   "version": 1,
+  "serialNumber": "urn:uuid:22b0cd8e-7894-443b-bc19-ffd4e2e1bf6e",
   "metadata": {
-    "timestamp": "1970-01-01T00:00:00.000000000Z",
+    "timestamp": "2026-04-08T15:31:40Z",
     "tools": [
       {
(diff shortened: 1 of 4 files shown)


mattkram (approved):
Awesome!

PR #111, opened 2026-04-14
fix: Use git commit timestamps for CI lockfile and SBOM freshness checks

mcg1969:
## Summary

- Fix spurious `Cargo.lock` regeneration in CI pre-commit that produced bot commits
- File timestamps are unreliable after `actions/checkout`, so `cargo generate-lockfile` was running unnecessarily and resolving to newer patch versions
- In CI, use `git log` commit timestamps to determine freshness:
  - `Cargo.toml` vs `Cargo.lock` for lockfile regeneration
  - `Cargo.lock` vs `SBOM.json` for SBOM regeneration
- Add `fetch-depth: 0` to checkout step so full history is available
- Local pre-commit behavior unchanged (file timestamp heuristic)

## Test plan
- [x] shellcheck passes
- [x] CI pre-commit job skips lockfile regeneration when Cargo.toml is unchanged
- [x] CI pre-commit job skips SBOM regeneration when SBOM.json is up to date

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -32,6 +32,7 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           token: ${{ secrets.ANACONDA_BOT_TOKEN }}
+          fetch-depth: 0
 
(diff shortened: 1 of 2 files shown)


mattkram (approved):
Thanks for the fix!

PR #120, opened 2026-04-17
fix: Add Windows support to install script

mattkram:
## Summary

The install script now recognizes MINGW/MSYS/Cygwin environments as Windows and downloads the correct `ana-windows-x86_64.exe` binary.

We need to do this in order to allow CI to us the install script on Windows runners, which run a form of MINGW and bash on Windows.

Changes:
- detect_os() recognizes MINGW*, MSYS*, and CYGWIN* as Windows
- map_target() includes windows-x86_64 target
- Binary name uses .exe suffix on Windows
- Tests now run on Windows with appropriate platform handling


jnoller (approved):
:shipit:

PR #129, opened 2026-04-21
feat: Improved UX for auth commands

mattkram:
## Summary

- Adds a unified `ui` module with `status` and `styles` submodules for consistent CLI output
- Improves `ana login` to show user info and token expiration after successful authentication
- Improves `ana whoami` with styled output showing account, subscriptions, and token info
- Adds `--json` flag to `ana whoami` for machine-readable output
- Fixes keyring to use same location as anaconda-auth (`~/.anaconda/keyring`)
- Fixes handling of empty keyring files
- Adds login/logout state checks with user prompts

## UI Module

New `ui::status` functions for consistent output:
- `success()` - green checkmark (✓)
- `great_success()` - sparkle emoji (✨) for final completion messages
- `error()` - red "error:" prefix
- `warn()` - amber exclamation (!)
- `info()` - plain text
- `waiting()` - dim text for in-progress states
- `highlight()` - blue text for values
- `dim()` - gray text for secondary info

## `ana whoami` Output

```
ACCOUNT
  name        Alice Smith
  username    asmith
  email       alice@example.com

SUBSCRIPTIONS
  Org Name (security)         2035-01-02 (8 years, 258 days)
  Another Org (commercial)    2027-01-01 (255 days)

TOKEN
  value       eyJ_****************************
  storage     /Users/user/.anaconda/keyring
```

- Shows all organizations with their subscription types and expiration dates
- Formats duration as "X years, Y days" for longer periods
- Shows "expired" in red for past expiration dates
- Token is masked, storage shows actual keyring path

## Login/Logout Improvements

- `ana login` now prompts "Login again? [y/N]" if already logged in
- `ana logout` now warns "Not logged in to {domain}" if already logged out
- Logout message includes full URL with https:// prefix for token revocation

## Example

Here is an example of the updated styling across the auth command lifecycle:

image

## Test plan

- [ ] Run `ana login` - should show styled output, user info, and token expiration after auth
- [ ] Run `ana login` when already logged in - should prompt before re-authenticating
- [ ] Run `ana logout` - should show styled confirmation with revocation URL
- [ ] Run `ana logout` when already logged out - should warn and exit gracefully
- [ ] Run `ana whoami` - should show styled account/subscription/token info
- [ ] Run `ana whoami --json` - should output raw JSON
- [ ] Run `ana whoami` when logged out - should show error with login hint

### DIFF
--- a/src/auth/actions.rs
+++ b/src/auth/actions.rs
@@ -462,10 +462,10 @@ pub fn logout() -> Result<(), AuthError> {
         "Logged out of {}",
         status::highlight(&config.domain)
     ));
-    status::success("Token removed from system keyring");
+    status::success("API key removed from system keyring");
     status::warn(&format!(
         "To fully revoke your token visit {}",
-        status::highlight(&format!("{}/settings/tokens", config.domain))
+        status::highlight(&format!("{}/app/profile/api-keys", config.domain))
     ));
     Ok(())
 }
--- a/src/auth/actions.rs
+++ b/src/auth/actions.rs
@@ -122,6 +122,95 @@ struct TokenErrorResponse {
     error_description: Option<String>,
 }
 
+/// Account information from the API.
+#[derive(Debug, Deserialize)]
+struct AccountInfo {
+    email: Option<String>,
+    user: Option<UserInfo>,
+}
+
+/// User information nested in account response.
+#[derive(Debug, Deserialize)]
+struct UserInfo {
+    username: Option<String>,
+}
+
+/// API key information from the API.
+#[derive(Debug, Deserialize)]
+struct ApiKeyInfo {
+    expires_at: Option<String>,
+}
+
+/// Print logged-in user status line.
+///
+/// Example: `✓ Logged in as kford@anaconda.com (anaconda)`
+fn print_logged_in_status(email: &str, org: &str) {
+    status::success(&format!(
+        "Logged in as {} ({})",
+        status::highlight(email),
+        org
+    ));
+}
+
+/// Calculate days from today until a date string (YYYY-MM-DD format).
+fn days_until_date(date_str: &str) -> Option<i64> {
+    let expires = chrono::NaiveDate::parse_from_str(date_str, "%Y-%m-%d").ok()?;
+    let today = chrono::Utc::now().date_naive();
+    Some((expires - today).num_days())
+}
+
+/// Print token expiration info.
+///
+/// Example: `  expires    2026-04-09 (365 days)`
+fn print_expiration(expires_at: &str) {
+    // Parse the expiration date and calculate days remaining
+    if let Some(days_remaining) = days_until_date(&expires_at[..10]) {
+        let days_str = if days_remaining == 1 {
+            "1 day".to_string()
+        } else {
+            format!("{} days", days_remaining)
+        };
+        eprintln!(
+            "  {}{}{}",
+            status::dim("expires    "),
+            status::highlight(&expires_at[..10]),
+            status::dim(&format!(" ({days_str})"))
+        );
+    }
+}
+
+/// Combined login information for display.
+struct LoginInfo {
+    email: String,
+    org: String,
+    expires_at: Option<String>,
+}
+
+/// Fetch login info (account + API key expiration) for display after login.
+async fn fetch_login_info(config: &Config) -> Result<LoginInfo, AuthError> {
+    let client = ApiClient::new()?;
+
+    // Fetch account info
+    let account_response = client.get("/api/account").send().await?;
+    let account: AccountInfo = account_response.json().await?;
+
+    let email = account
+        .email
+        .or_else(|| account.user.and_then(|u| u.username))
+        .unwrap_or_else(|| "unknown".to_string());
+
+    // Fetch API key info for expiration
+    let keys_response = client.get("/api/auth/api-keys").send().await?;
+    let keys: Vec<ApiKeyInfo> = keys_response.json().await.unwrap_or_default();
+    let expires_at = keys.first().and_then(|k| k.expires_at.clone());
+
+    Ok(LoginInfo {
+        email,
+        org: config.domain.clone(),
+        expires_at,
+    })
+}
+
 /// Perform the device authorization flow.
 pub async fn login() -> Result<(), AuthError> {
     // We use a new, unauthenticated client instead of ApiClient, since
@@ -277,7 +366,16 @@ pub async fn login() -> Result<(), AuthError> {
 
             // Save to keyring
             save_api_key(&config, &api_key)?;
-            status::success("Token stored in system keyring");
+            status::success("Token stored in keyring");
+
+            // Fetch and display user info
+            if let Ok(login_info) = fetch_login_info(&config).await {
+                print_logged_in_status(&login_info.email, &login_info.org);
+                if let Some(ref expires_at) = login_info.expires_at {
+                    print_expiration(expires_at);
+                }
+            }
+
             return Ok(());
         }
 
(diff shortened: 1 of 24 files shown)


src/auth/actions.rs
```diff
@@ -90,35 +86,99 @@ impl ApiClient {
...
 
-/// OpenID Connect discovery document.
-#[derive(Debug, Deserialize)]
-struct OpenIdConfig {
-    device_authorization_endpoint: Option<String>,
-    token_endpoint: String,
+/// Print logged-in user status line.
+///
+/// Example: `✓ Logged in as user@example.com`
+fn print_logged_in_status(email: &str) {
```
zfralish, line 92: I think all these helper function may make more sense under the ui dir, as like `helpers.rs` or `format.rs`

src/auth/responses.rs
zfralish, line 1: This might warrant a `/schema` directory as we need more of these.

src/help/styles.rs
zfralish, line 1: I would like something like

```
-src
--ui
---styles
----help.rs
```

zfralish (approved):
Overall, love this. Left a few nits, disregard if not of concern.

No need to deal with this now but the file structure is becoming very confusing. I would like to take some time to visit a scalable and discoverable file structure.

My ideal would be something like each commands having its own file under a commands directory, I know rust is more fussy about file structure though.

Then I would like to compose functional calls into services, like `api_key_service.rs`

```
-src
--cmds
---login_cmd.rs <---- imports services
--services
---api_key_service.rs
```

This may not be a correct structure in rust though 🤷🏻

src/help/styles.rs
mattkram: I like this. I will merge this to get the code in main, and follow up with a pure refactor for all the structural changes. You are right, in that I think we've grown to a point now where that amount of structure is warranted.

tests/conftest.py
```diff
@@ -12,6 +12,66 @@
 from mock_auth_server import MockAuthServer
 
 
+def assert_output_contains(text: str, *patterns: str) -> None:
```
marcoesters: Typically, `conftest` is reserved for fixtures and plug-ins. We have enough stuff now (like the `AnaRunner` type definition, `IS_WINDOWS`, etc.) to use a helper module.

tests/conftest.py
```diff
@@ -12,6 +12,66 @@
...
+            msg = (
+                f"Pattern not found: {pattern!r}\n\n"
+                f"Already matched:{matched_str}\n\n"
+                f"Searching from line {current_line + 1}:\n{context}\n\n"
+                f"Full output:\n{text}"
+            )
+            raise AssertionError(msg)
+
+        matched.append(pattern)
+        pos = idx + len(pattern)
```
marcoesters: I find this a little difficult to read because there is so much code after the if-condition. Consider doing this instead:

```
if idx >= 0:
    matched.append(pattern)
    pos = idx + len(pattern)
    continue
<error handling code here>
```

src/config.rs
```diff
@@ -220,9 +220,9 @@ impl Config {
     }
 }
 
-/// Get the default keyring path (~/.ana/keyring or $ANA_HOME/keyring).
+/// Get the default keyring path (~/.anaconda/keyring or $ANA_KEYRING_PATH).
```
marcoesters: The code isn't looking at `$ANA_KEYRING_PATH`, so the comment or the code needs to be updated.

src/ui/styles.rs
```diff
@@ -0,0 +1,97 @@
...
+    BoxBorder,
+    BoxText,
+}
+
+impl UiColor {
+    /// Get the hex value for this color.
+    fn hex(&self) -> &'static str {
+        match self {
+            Self::Green => "#3fb950",
+            Self::Red => "#f85149",
```
marcoesters, line 26: Just a note: we should check with UX to ensure that these colors are accessible for everyone.

src/ui/status.rs
```diff
@@ -0,0 +1,78 @@
...
+pub fn success(msg: &str) {
+    eprintln!("{} {}", UiColor::Green.apply_to("✓"), msg);
+}
+
+/// Print an exciting success message with sparkles.
+///
+/// Use for final completion messages. Example output: `✨ You can now install packages!`
+pub fn great_success(msg: &str) {
+    eprintln!("{} {}", UiColor::Green.apply_to("✨"), msg);
+}
```
marcoesters: This function is not used anywhere, which leads to compiler warnings. I'm also not sure we need to distinguish between `success` and `great_success`.

src/auth/actions.rs
```diff
@@ -90,35 +86,99 @@ impl ApiClient {
...
+/// - 400 days -> ("1 year, 35 days", false)
+/// - 1 day -> ("1 day", false)
+/// - -30 days -> ("expired", true)
+fn format_duration(days: i64) -> (String, bool) {
+    if days < 0 {
+        return ("expired".to_string(), true);
+    }
+
+    let years = days / 365;
+    let remaining_days = days % 365;
```
marcoesters: That won't work for leap years. We should consider using a library instead that does the parsing for us.

src/auth/actions.rs
```diff
@@ -90,35 +86,99 @@ impl ApiClient {
...
+        } else if years == 1 {
+            format!("1 year, {} days", remaining_days)
+        } else {
+            format!("{} years, {} days", years, remaining_days)
+        }
+    } else if days == 1 {
+        "1 day".to_string()
+    } else {
+        format!("{} days", days)
+    };
```
marcoesters, line 143:
```suggestion
    let day_unit = if remaining_days == 1 { "day" } else { "days" };
    let year_unit = if years == 1 { "year" } else { "years" };
    let s = if years > 0 {
        if remaining_days == 0 {
            format!("{} {}", years, year_unit);
        } else {
            format!("{} {}, {} {}", years, year_unit, remaining_days, day_unit)
        }
    } else {
        format("{} {}", remaining_days, day_unit)
    };
```

This is more compact and (in my opinion) more legible.

src/auth/actions.rs
```diff
@@ -2,32 +2,34 @@
...
 use crate::http::{Client, bearer_header, build_client};
 use crate::input::KeyListener;
+use crate::ui::status;
 
 const REQUEST_TIMEOUT: Duration = Duration::from_secs(30);
 
 /// Print a QR code to the terminal with indentation.
 fn print_qr(qr: &str) {
-    println!();
+    status::blank_line();
```
marcoesters: Maybe use `eprintln!` for consistency? Or do we want to route all our printing through helper functions?

src/ui/status.rs
mattkram: This came in from an upcoming PR that will build on top of this.

The idea is to have a final message style. I'll pull from here and add it in the follow up.

tests/conftest.py
mattkram: Good call. I had the exact same thought and didn't pull the trigger, but since you detected the same code smell it's worth refactoring.

src/ui/styles.rs
mattkram: Agree. We will also likely need to add some contingency for simpler terminals that don't support TrueColor in the future. For now, will merge as is, but I'll make an issue around visual issues for tracking.

marcoesters: Terminal support is one aspect, but I was actually mostly thinking about color vision deficiencies here, especially with red and green.

src/ui/status.rs
mattkram: Removed it for now.

src/auth/actions.rs
mattkram: Agree, great feedback.

src/auth/actions.rs
mattkram: Refactored to use the `chrono` library we added.

PR #130, opened 2026-04-21
fix: Use shim instead of symlinks on Windows

marcoesters:
## Summary

Windows does not support symlinks by default and require developer mode to be enabled (see also the [rust documentation](https://doc.rust-lang.org/stable/std/os/windows/fs/fn.symlink_file.html#limitations)). This requires administrator privileges, which not every user has. Hard links are not an appropriate alternative because they do not work across drives and require regeneration whenever the original binary updates.

Instead, use a compiled shim that just forwards the call to the appropriate binary. Shimming is a common way to solve the symlink problem on Windows and is deployed by tools like chocolatey or even `conda` (the `.exe` entry points in `Scripts` just call the `.py` file of the same name). I tested this with in PowerShell, Cygwin, MSYS2, cmd.exe, and Git Bash.

The shim is statically compiled and embedded into the `ana` binary. When a tool is installed, that binary blob will be written into `bin` with the name of binary it shims (i.e., the shim for `pixi` is `bin\pixi.exe`). When called, the shim does two things:

1. It reads the `shims.cfg` file and find the entry that corresponds to the binary name. For pixi, it will find the line that contains `pixi=`. The value is the file path to the actual binary relative to the `tools` directory.
1. It then calls that binary and forwards all command line arguments to the function call.

`shims.cfg` is created at runtime and updated every time a tool is installed or uninstalled.

This PR also fixes a few other Windows errors and issues:

* Use the `anaconda.exe` binary inside `Scripts`, not `bin`. This required a refactor for how we store `binaries` in the `Tool` object.
* Add instructions for how to set the GitHub token in Windows-native formats.
* Several integration test fixes:
  * Patch `USERPROFILE` instead of `HOME` and also set the rattler cache directory in integration tests.
  * Ensure that UTF-8 encoding is used in subprocess calls or non-ASCII characters aren't rendered properly (Windows uses UTF-16).
  * Skip shell tests outside of GitBash
  * Run integration tests in PowerShell to ensure they are compatible with native Windows shells
  * Use Windows-appropriate conventions in integration tests (backslashes in path strings, `.exe` file endings for binaries)

## Ticket(s)
Jira: [CLI-492](https://anaconda.atlassian.net/browse/CLI-492)

## AI disclosure

Claude was used to help create the shim code, answer general rust questions, and debug. Claude wrote the unit tests in the rust code.

[CLI-492]: https://anaconda.atlassian.net/browse/CLI-492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/tools/install.rs
+++ b/src/tools/install.rs
@@ -1,6 +1,6 @@
 //! Package installation from lockfiles via rattler.
 
-use std::{path::Path, str::FromStr, time::Instant};
+use std::{path::Path, path::PathBuf, str::FromStr, time::Instant};
 
 use indicatif::{MultiProgress, ProgressDrawTarget};
 use miette::{Context, IntoDiagnostic};
@@ -29,14 +29,14 @@ pub async fn install_tool(name: &str) -> miette::Result<()> {
     let lock_content =
         tools::content(name).ok_or_else(|| miette::miette!("unknown tool: {}", name))?;
 
-    let binaries = tools::binaries(name).unwrap_or(&[]);
+    let binaries = tools::binaries(name).unwrap_or(Vec::new());
 
     eprintln!("Installing {} into {}", name, prefix.display());
 
     install_from_lockfile(&prefix, &lock_content).await?;
 
     // Create symlinks in bin directory
-    create_bin_symlinks(&prefix, binaries)?;
+    create_bin_symlinks(&prefix, &binaries)?;
 
     // Tool-specific post-install configuration
     if name == "pixi" {
@@ -122,7 +122,7 @@ pub async fn install_from_lockfile(prefix: &Path, lock_content: &str) -> miette:
 }
 
 /// Create symlinks for the tool's binaries in ~/.ana/bin/
-fn create_bin_symlinks(prefix: &Path, binaries: &[&str]) -> miette::Result<()> {
+fn create_bin_symlinks(prefix: &Path, binaries: &[PathBuf]) -> miette::Result<()> {
     let bin_dir = paths::bin_dir();
     std::fs::create_dir_all(&bin_dir)
         .into_diagnostic()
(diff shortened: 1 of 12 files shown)


tests/conftest.py
```diff
@@ -22,6 +23,7 @@ def _find_repo_root() -> Path:
 
 
 REPO_ROOT = _find_repo_root()
+IS_WINDOWS = sys.platform == "win32"
```
mattkram, line 26: Should we import this elsewhere? May be over-engineering, not sure.

src/update.rs
```diff
@@ -44,7 +44,20 @@ impl std::fmt::Display for Error {
...
+                #[cfg(windows)]
+                {
+                    writeln!(f, " Run:")?;
+                    writeln!(f, "   PowerShell: $env:GITHUB_TOKEN=(gh auth token)")?;
+                    writeln!(
+                        f,
+                        "   cmd.exe: for /f %i in ('gh auth token') do set GITHUB_TOKEN=%i"
+                    )?;
+                    writeln!(f, "   git bash: export GITHUB_TOKEN=$(gh auth token)")
+                }
```
mattkram, line 60: Cool! This will go away once we are public, or have a static hosting solution.

src/tools/install.rs
```diff
@@ -181,6 +174,103 @@ fn create_bin_symlink(bin_dir: &Path, prefix: &Path, binary: &str) -> miette::Re
     Ok(())
 }
 
+#[cfg(windows)]
+/// Embedded shim binary (compiled from src/bin/shim.rs)
+const SHIM_BINARY: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/shim.exe"));
```
mattkram, line 179: Clever!

mattkram (approved):
This is really great! I missed at first that the `shims.cfg` was dynamically created/managed. It also wasn't initially clear to me that we were going to compile in the shim binary, but that is clearer to me now. It may be worthwhile to just review the PR description and make sure those design details are captured.

tests/conftest.py
marcoesters: I will probably use the `conda` work to do some refactoring of our test helper functions. `REPO_ROOT` and `IS_WINDOWS` should go in there, too.

PR #131, opened 2026-04-21
feat: Add CommandContext for telemetry attributes

zfralish:
## Summary

Introduces a `CommandContext` struct that carries cross-cutting concerns through the command call stack, inspired by Go's `context.Context` pattern.

### The Pattern

In Go, it's common to pass a `context.Context` as the first parameter to functions that need access to request-scoped values, cancellation signals, or deadlines. This PR adopts a similar approach for ana-cli:

```rust
pub struct CommandContext {
    pub telemetry: TelemetryContext,
    // Future: config, auth_state, etc.
}
```

Commands receive `&mut CommandContext` and can add telemetry attributes at any depth in the call stack:

```rust
pub async fn install_tool(name: &str, ctx: &mut CommandContext) -> miette::Result<()> {
    ctx.telemetry.add("tool_name", name.to_string());
    // ...
}
```

### Why This Pattern?

- **Extensibility**: New cross-cutting concerns (config, auth state, cancellation) can be added to `CommandContext` without changing function signatures everywhere
- **Telemetry flexibility**: Commands can add attributes from anywhere in the call stack, not just at the top level
- **Clean separation**: Context lives in its own module (`src/context.rs`), keeping `cli.rs` focused on command dispatch

### Changes

- Add `src/context.rs` with `TelemetryContext` and `CommandContext`
- Pass `&mut CommandContext` to all command functions
- Tool install/uninstall now capture `tool_name` attribute automatically
- Common attributes (`os`, `arch`, `version`, `command`) are set automatically

## Test plan

- [x] `cargo check` passes
- [x] `cargo test` passes (80 tests)

### DIFF
--- a/src/feedback.rs
+++ b/src/feedback.rs
@@ -1,3 +1,5 @@
+use crate::context::CommandContext;
+
 const FEEDBACK_BASE_URL: &str = "https://docs.google.com/forms/d/e/1FAIpQLSeGd9p7pQSHvjIc6RNShjTQCGmM-5_3xkPNpNfYk102-HZB8Q/viewform";
 
 /// Feedback type for the feedback form
@@ -17,7 +19,7 @@ pub fn parse_feedback_type(bug: bool, feature: bool) -> Option<FeedbackType> {
     }
 }
 
-pub fn open_feedback(feedback_type: Option<FeedbackType>, description: Option<String>) {
+pub fn open_feedback(feedback_type: Option<FeedbackType>, description: Option<String>, _ctx: &mut CommandContext) {
     let mut params = vec![("usp", "pp_url".to_string())];
 
     if let Some(ft) = feedback_type {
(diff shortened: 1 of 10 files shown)


src/feedback.rs
```diff
@@ -17,7 +19,11 @@ pub fn parse_feedback_type(bug: bool, feature: bool) -> Option<FeedbackType> {
     }
 }
 
-pub fn open_feedback(feedback_type: Option<FeedbackType>, description: Option<String>) {
+pub fn open_feedback(
+    feedback_type: Option<FeedbackType>,
+    description: Option<String>,
+    _ctx: &mut CommandContext,
```
mattkram: Do you think we should standardize on this being the first positional argument?

mattkram (approved):
I like it. Just wondering if we should conventionally always make it the first parameter.

Some of the upstream changes will need resolution prior to merge.

src/feedback.rs
zfralish: Yeah, I think that is a good idea, I will make these changes today.

PR #136, opened 2026-04-23
feat: Add anaconda-anon-usage telemetry tokens to user-agent string

zfralish:
## Summary

Integrates the [`anaconda-anon-usage`](https://crates.io/crates/anaconda-anon-usage) Rust crate to add AAU identity tokens to the user-agent string for anonymous usage telemetry.

Based on: https://github.com/Anaconda-Sandbox/ana-cli-mcg/pull/8

- Adds `anaconda-anon-usage` dependency for token generation
- Rewrites `src/ua/mod.rs` to use the crate  
- Adds `ana self user-agent` command with optional `--prefix` flag
- Flushes deferred token writes at process exit
- Adds tests for header injection

### User-Agent Format

```
ana/0.0.6 Darwin/25.2.0 OSX/26.2 rattler/0.40.3 aau/0.8.0 c/... s/... e/...
```

### Token Types

| Prefix | Name | Persistence |
|--------|------|-------------|
| `c/` | Client | MAC-tied, saved to `~/.conda/aau_token` |
| `s/` | Session | Random per-process |
| `e/` | Environment | Per-conda-env |
| `a/` | Anaconda Cloud | Derived from auth JWT |

## Test plan

- [x] `cargo test` — 100 tests pass
- [x] New tests verify AAU tokens are injected into HTTP headers
- [ ] Run `ana self user-agent` and confirm token output


Cargo.toml
```diff
@@ -14,6 +14,7 @@ diagnostics = ["dep:sentry"]
 feedback = ["dep:urlencoding"]
 
 [dependencies]
+anaconda-anon-usage = "0.8.0-pre.3"
```
mcg1969: we're up to 0.8.0-pre.7 now. But we should get to 0.8.0 very soon

src/ua/mod.rs
```diff
@@ -1,52 +1,109 @@
...
+    }
+}
+
+/// Return the base user-agent string (without AAU tokens).
+fn base_user_agent() -> &'static str {
+    static USER_AGENT: LazyLock<String> = LazyLock::new(|| {
+        format!(
+            "ana/{} {} rattler/{}",
+            VERSION,
+            platform::platform_string(),
```
mcg1969: You have duplicate platform strings now (see `ana self user-agent` to see this). The reason is that in order to make it easier to use the crate in other contexts, I gave anaconda-anon-usage the ability to supply the platform string. So we can actually remove `platform::platform_string` altogether, delegating it entirely to the crate, or we can disable the platform string generator in anaconda-anon-usage. I vote the first option since the crate will have the code regardless

zfralish: Yep agree, will remove.

src/ua/mod.rs
```diff
@@ -1,52 +1,109 @@
...
+            None
+        }
+    };
+    anaconda_anon_usage::Config {
+        env_prefix: env_prefix().map(|s| s.to_string()),
+        anaconda_jwt: jwt,
+        prefix: Some(base_user_agent().to_string()),
+        platform: true,
+        rattler_version: Some(RATTLER_VERSION.to_string()),
+        reqwest_version: None,
```
mcg1969, line 54: I recommend populating reqwest_version here if possible. One mechanism you can use, for both reqwest_version and rattler_version, is to populate constants using build.rs, and have build.rs harvest the version numbers from Cargo

mcg1969 (approved):

PR #139, opened 2026-04-23
feat: Add PowerShell installer

marcoesters:
## Summary

Add a PowerShell installer for `ana`.  The installer provides the same functionality as the `sh` installer. The script uses PowerShell 5 because it is the default PowerShell version on Windows. For macOS and Linux, PowerShell hasn't been available until version 6.

## Testing

Refactor the tests so that the same mock installer fixtures that are used in SH are available for PowerShell. To make the binary tests work on Windows, I had to resort to a hack: the mock script will be called `.exe`, but is then renamed in the tests to run `.cmd`. Windows refuses to run `.exe` files when they are not PE files, but `.cmd` files can be run natively in `cmd.exe`, PowerShell, and Git Bash.

I also had to change the shell on Windows to `powershell` instead of `pwsh` because calling `powershell` from inside `pwsh` made `Get-FileHash` unavailable even though it has been supported since PowerShell 4.

## Documentation

Update the documentation to include PowerShell installation instructions.

Jira: [CLI-399](https://anaconda.atlassian.net/browse/CLI-399)

[CLI-399]: https://anaconda.atlassian.net/browse/CLI-399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/install.ps1
```diff
@@ -0,0 +1,453 @@
...
+      ANA_NO_PATH_UPDATE       Set to non-empty to skip PATH update
+      ANA_BOOTSTRAP            Set to "false" to skip bootstrap
+      ANA_FORCE_INSTALL        Set to non-empty to overwrite without prompting
+      GITHUB_TOKEN             Same as -Token
+
+.PARAMETER InstallDir
+    Installation directory (default: ${env:USERPROFILE}\.local\bin (Windows) or ${env:HOME}/.local/bin).
+
+.PARAMETER Version
+    Version to installi (default: latest).
```
mattkram:
```suggestion
    Version to install (default: latest).
```

mattkram (dismissed):
This is really great! I don't have a great way to manually test it myself, but I recommend we merge it and ask QA to check it on the Windows boxes.

Since it's still draft, feel free to dismiss and ask for another review if you want. Up to you on when to merge.

mattkram (approved):
Looks great to me. I think we should merge this and ask QA to validate the installation across platforms.

PR #142, opened 2026-04-27
feat: Add `api fetch` command for making HTTP requests

zfralish:
## Summary
- Adds new `ana api fetch` command for making HTTP requests to the configured API base URL
- Supports multiple HTTP methods (GET, POST, PUT, PATCH, DELETE)
- Includes options for query params (`-q`), raw body (`-d`), and JSON body (`-j`)
- Adds HTTP client to `CommandContext` for shared use across commands

## Test plan
- [ ] Run `ana api fetch /api/iam/whoami` - should make GET request
- [ ] Run `ana api fetch /api/test --method POST -j '{"foo":"bar"}'` - should POST with JSON body
- [ ] Run `ana api fetch /api/test -q key=value,other=123` - should include query params
- [ ] Verify `--data` and `--json` flags conflict with each other

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -10,9 +10,9 @@ use crate::auth;
 use crate::config::{self, Config};
 use crate::context::CommandContext;
 use crate::feature;
-use crate::fetch::api_fetch;
 #[cfg(feature = "feedback")]
 use crate::feedback::{self, FeedbackType};
+use crate::fetch::api_fetch;
 use crate::help;
 use crate::tools;
 use crate::update;
@@ -780,5 +780,4 @@ mod tests {
             missing
         );
     }
-
 }
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -509,9 +509,9 @@ async fn api_fetch(
         request = request.body(body.to_string());
     }
     if let Some(body) = json {
-        request = request
-            .header("Content-Type", "application/json")
-            .body(body.to_string());
+        let parsed: serde_json::Value = serde_json::from_str(body)
+            .map_err(|e| format!("Invalid JSON: {}", e))?;
+        request = request.json(&parsed);
     }
     let response = request.send().await?;
     let status = response.status();
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -123,6 +123,13 @@ pub enum Action {
         force: bool,
     },
     ToolList,
+    ApiFetch {
+        method: String,
+        url: String,
+        query_args: Option<String>,
+        data: Option<String>,
+        json: Option<String>,
+    },
 }
 
 impl Action {
@@ -147,6 +154,7 @@ impl Action {
             Action::ToolInstall { .. } => "tool.install",
             Action::ToolUninstall { .. } => "tool.uninstall",
             Action::ToolList => "tool.list",
+            Action::ApiFetch { .. } => "api.fetch",
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/cli.rs
```diff
@@ -398,6 +439,46 @@ fn get_subcommand(name: &str) -> clap::Command {
...
+        request = request.query(&pairs);
+    }
+    if let Some(body) = data {
+        request = request.body(body.to_string());
+    }
+    if let Some(body) = json {
+        request = request
+            .header("Content-Type", "application/json")
+            .body(body.to_string());
+    }
```
mattkram: I'm wondering if we should do any client-side parsing here first and use the .json method. That would allow immediate feedback on poorly constructed json strings.

src/context.rs
```diff
@@ -43,17 +46,25 @@ impl TelemetryContext {
...
     pub fn new() -> Self {
+        let config = Config::load();
+        let mut builder = reqwest::Client::builder();
+        if let Ok(Some(api_key)) = auth::get_api_key(&config) {
+            builder = builder.default_headers(bearer_header(&api_key));
+        }
+        let client = Client::new(builder, config.base_url()).expect("failed to create HTTP client");
         Self {
             telemetry: TelemetryContext::new(),
+            client,
```
mattkram: I would assume we already have a helper function to do this generally for our http client. We should try to only do it in one place.

zfralish: Done.

src/cli.rs
zfralish: Done - now parsing with serde_json::from_str and using .json() method for immediate validation feedback.

src/cli.rs
```diff
@@ -440,6 +481,46 @@ fn get_subcommand(name: &str) -> clap::Command {
...
 }
 
+async fn api_fetch(
+    ctx: &mut CommandContext,
+    method: &str,
+    url: &str,
+    query_args: Option<&str>,
+    data: Option<&str>,
+    json: Option<&str>,
+) -> Result<(), Box<dyn std::error::Error>> {
```
mattkram: It might be good to put this into its own module.

Additionally, I think we should enforce required authentication when using the `ana api fetch` command, to prevent abuse from anonymous actors.

src/cli.rs
```diff
@@ -440,6 +481,46 @@ fn get_subcommand(name: &str) -> clap::Command {
...
+    let method_upper = method.to_uppercase();
+    let mut request = match method_upper.as_str() {
+        "GET" => ctx.client.get(url),
+        "POST" => ctx.client.post(url),
+        "PUT" => ctx.client.put(url),
+        "PATCH" => ctx.client.patch(url),
+        "DELETE" => ctx.client.delete(url),
+        _ => return Err(format!("Unsupported HTTP method: {}", method).into()),
+    };
```
mattkram: It might be helpful to add some other header indicating it's a "raw" HTTP request. Not sure if we would need it, but it could be a helpful emergency filter to differentiate "a bunch of bots figured out how to use `ana api fetch`" from "real" requests.

src/cli.rs
zfralish: Done

src/cli.rs
zfralish: Done

mattkram (approved):
Awesome!

PR #153, opened 2026-04-28
feat: Add static hosting support for self-update

mattkram:
## Summary
- Adds `ANA_SELF_UPDATE_URL` env var to configure the update source
- Default: `https://anaconda.sh` (static hosting)
- Set to `github` to use GitHub Releases (requires `GITHUB_TOKEN`)
- Fetches manifest from `/releases.json` and downloads binaries from `/releases/{channel}/{version}/ana-{target}`
- No authentication required for static hosting
- Passes Cloudflare Zero Trust headers (`CF-Access-Client-Id`, `CF-Access-Client-Secret`) when available for CI access

Styling updates are deferred to a follow-up PR.

## Demonstration

image

## Test plan
- [x] `ana self update --list` works with static hosting (no token needed)
- [x] `ANA_PRERELEASES=true ana self update --list` shows dev channel
- [x] `ANA_SELF_UPDATE_URL=github` falls back to GitHub Releases
- [ ] `ana self update` successfully downloads and replaces binary


src/auth/keyring.rs
```diff
@@ -166,6 +166,7 @@ mod tests {
             metrics_console_exporter: false,
             metrics_skip_internet_check: true,
             include_prereleases: false,
+            self_update_url: Some("https://example.com".to_string()),
```
zfralish, line 169: want to keep this?

zfralish (approved):

src/auth/keyring.rs
mattkram: I think this is fine in a test module.

PR #164, opened 2026-04-30
fix: Get vault secrets in a job running on self hosted runner

anaconda-zday:


mattkram (dismissed):

mattkram (approved):
The bug is mine. I will meet and fix in my branch.

PR #168, opened 2026-05-04
feat: Implement better styling for `ana self update`

mattkram:
## Summary

Implements the designs for `ana self update`. The essential behavior is the same, however the terminal output and user interfaces have been overhauled.

### Key changes

- **Removed `--yes/-y` flag** - updates now run automatically without confirmation prompt
- **Improved help output** - better formatting with examples, `[OPTIONS]` before positional args
- **Styled `--check` output** - shows update availability with version comparison and relative time
- **Styled `--list` output** - shows all available versions with dates (when available), highlights "Latest" and "Current" versions
- **Styled download progress** - cleaner progress bar with elapsed time
- **Unified success output** - consistent "UPDATED" message across all update paths with timing info
- **GitHub API pagination** - now fetches up to 100 releases (was 30) to show all stable versions

### `ana self update --help`

image

## `ana self update --check`

image

## `ana self update --list`

image

## `ana self update`

After success:

image

During download:

image

## Test plan

- [x] `ana self update --help` shows styled help with examples
- [x] `ana self update --list` shows versions with dates aligned (GitHub) or without dates (static hosting)
- [x] `ana self update --check` shows styled update availability
- [x] `ana self update` downloads and shows success message with timing
- [x] All existing update tests pass

Jira: [CLI-573](https://anaconda.atlassian.net/browse/CLI-573)

[CLI-573]: https://anaconda.atlassian.net/browse/CLI-573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/help/term.rs
+++ b/src/help/term.rs
@@ -25,6 +25,35 @@ fn print_command_row(term: &Term, name: &str, desc: &str) {
     ));
 }
 
+/// Print an option row with optional short/long flags
+fn print_option_row(term: &Term, short: Option<&str>, long: Option<&str>, desc: &str) {
+    let cmd_style = HelpStyle::Command.style();
+    let dim_style = HelpStyle::Dim.style();
+
+    let (name, visible_len) = match (short, long) {
+        (Some(s), Some(l)) => {
+            let styled = format!(
+                "{}{}{}",
+                cmd_style.apply_to(s),
+                dim_style.apply_to(", "),
+                cmd_style.apply_to(l)
+            );
+            let visible = s.len() + 2 + l.len(); // "s" + ", " + "l"
+            (styled, visible)
+        }
+        (Some(s), None) => (cmd_style.apply_to(s).to_string(), s.len()),
+        (None, Some(l)) => (cmd_style.apply_to(l).to_string(), l.len()),
+        (None, None) => (String::new(), 0),
+    };
+
+    let padding = " ".repeat(20_usize.saturating_sub(visible_len));
+    let styled_desc = HelpStyle::Desc.style().apply_to(desc);
+    let _ = term.write_line(&format!(
+        "{}  {name}{padding} {styled_desc}",
+        left_margin()
+    ));
+}
+
 /// Print a section header
 fn print_section(term: &Term, name: &str) {
     let _ = term.write_line(&format!(
@@ -186,13 +215,14 @@ fn print_section_blocks(term: &Term, subcommands: &HashMap<String, String>) {
 
 fn print_options_block(term: &Term) {
     print_section(term, "OPTIONS");
-    print_command_row(
+    print_option_row(
         term,
-        "-v, --verbose",
+        Some("-v"),
+        Some("--verbose"),
         "Increase verbosity (can be repeated)",
     );
-    print_command_row(term, "-V, --version", "Show the ana version");
-    print_command_row(term, "-h, --help", "Show this message");
+    print_option_row(term, Some("-V"), Some("--version"), "Show the ana version");
+    print_option_row(term, Some("-h"), Some("--help"), "Show this message");
     let _ = term.write_line("");
 }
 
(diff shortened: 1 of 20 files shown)


src/help/term.rs
```diff
@@ -186,12 +190,39 @@ pub fn print_subcommand_help(cmd: &clap::Command) {
         let _ = term.write_line("");
     }
 
-    // Usage - render and ensure it starts with "ana "
-    let usage = cmd.clone().render_usage().to_string();
-    let usage = if usage.starts_with("Usage: ana ") {
-        usage
+    // Usage - build from path to ensure correct nesting
+    // Filter out built-in help/version flags (which have --long options)
+    let is_builtin_flag = |a: &clap::Arg| {
```
zfralish: oof this is hard to read

zfralish (approved):

src/help/term.rs
```diff
@@ -116,6 +118,7 @@ fn print_examples_block(term: &Term) {
         }
     }
 
+    // Bottom border
```
marcoesters:
```suggestion
```

Duplicate

src/help/term.rs
```diff
@@ -105,7 +107,7 @@ fn print_examples_block(term: &Term) {
         ));
 
         // Add spacing between examples (except after last)
-        if i < HELP_EXAMPLES.len() - 1 {
+        if i < examples.len() - 1 {
```
marcoesters, line 179: I think you can avoid the repeated length calculation by comparing references:

```
let last = examples.last();
for example in examples {
    // Command line (left_margined)
    ...
    // Add spacing between examples (except after last)
    if !std::ptr::eq(example, last.unwrap()) {
    }
}
```

Probably not too dramatic in terms of performance, but just an alternative.

src/help/term.rs
```diff
@@ -186,12 +190,39 @@ pub fn print_subcommand_help(cmd: &clap::Command) {
         let _ = term.write_line("");
     }
 
-    // Usage - render and ensure it starts with "ana "
-    let usage = cmd.clone().render_usage().to_string();
-    let usage = if usage.starts_with("Usage: ana ") {
-        usage
+    // Usage - build from path to ensure correct nesting
+    // Filter out built-in help/version flags (which have --long options)
```
marcoesters:
```suggestion
    // Filter out built-in help/version flags
```

I think that addition confused me more than it helped.

src/help/term.rs
```diff
@@ -186,12 +190,39 @@ pub fn print_subcommand_help(cmd: &clap::Command) {
...
+    let is_builtin_flag = |a: &clap::Arg| {
+        (a.get_id() == "help" || a.get_id() == "version")
+            && (a.get_long().is_some() || a.get_short().is_some())
```
marcoesters: Why is the `&&` necessary? Are there `help`/`version` subcommands that don't have long or short options?

src/update.rs
```diff
@@ -339,95 +362,277 @@ pub async fn check_update(current_version: &str) -> Result<UpdateCheck, Error> {
...
+    let duration = now.signed_duration_since(published);
+
+    let days = duration.num_days();
+    if days == 0 {
+        "released today".to_string()
+    } else if days == 1 {
+        "released yesterday".to_string()
+    } else if days < 30 {
+        format!("released {} days ago", days)
+    } else if days < 365 {
```
marcoesters: Since months and years do not have constant numbers of days, `TimeDelta` is not the right tool to do this accurately. `days`, `months`, and `years` will have to be calculated separately.

marcoesters (commented):
This seems to work on Windows, or at least in the Windows Terminal. I just have a few minor comments.

src/help/term.rs
mattkram: Agree, good call out. I refactored it for readability.

src/help/term.rs
mattkram: I refactored this a lot, so hopefully it's all more clear. This was basically saying it's built-in if it's help or version, and it has to define either long or short flag. That was probably superfluous.

src/update.rs
mattkram: I refactored this and added a note. It is okay to be approximate, but you are right to call it out, especially the "30 days in a month" assumption.

I looked into `chrono-humanize` but it didn't feel worth it for a dependency. Since this is just for display, I elected for the refactor and a bit less perfection. We can always re-evaluate.

src/help/term.rs
mattkram: For now, I will leave it. It's a cool solution, but it feels slightly less readable to me.

PR #178, opened 2026-05-05
feat: Remove pre-releases and speed up push events

marcoesters:
## Summary

Remove creating the pre-release for each merge. Currently, each push to `main` creates a pre-release, even if the changes are not related to the code base. This is unnecessary since artifacts are already available for each push to `main` and makes the releases page difficult to navigate. We could increase the retention time if there are concerns about accessing artifacts for a specific commit. The `conda` packages still exist as a fallback and are just a binary repack.

Additionally, I noticed that the current merging behavior runs through the CI three times:

1. During the merge group build
2. After pushing to `main`
3. When building the pre-release

While removing the pre-releases removes some of that inefficiency, the push event still contains some redundancy. To remove some of it, I decided to skip tests during push events. We still need to run the extra build to upload the `conda` package to the `dev` label because I couldn't find a way to add the upload step to the merge group while keeping the upload environment protected.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,12 +1,14 @@
 name: CI
 
 on:
-  push:
-    branches:
-      - main
   pull_request:
   merge_group:
   workflow_call:
+    inputs:
+      skip-tests:
+        description: Whether to skip tests
+        type: boolean
+        default: false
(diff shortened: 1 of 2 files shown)


mattkram (approved):
Nice, thanks!

PR #182, opened 2026-05-06
ci: Parallelize Windows CI jobs

mattkram:
## Summary
- Split Windows CI into parallel jobs to reduce total CI time
- Build job compiles binary + test harnesses, then parallel jobs run:
  - Unit tests (restore cargo cache)
  - Integration tests in bash (download binary artifact)
  - Integration tests in PowerShell (download binary artifact)

## Results

**Before (sequential):** Windows total ~30 min, CI completion ~30 min

| Job | Duration |
|-----|----------|
| Build (windows-latest) | 29m 26s |
| Build (ubuntu-latest) | 9m 50s |
| Build (macos-latest) | 4m 9s |
| Build (ubuntu-24.04-arm) | 3m 51s |

**After (parallel):** Windows build ~5 min + longest test job ~9 min, CI completion ~19 min

| Job | Duration |
|-----|----------|
| Build (windows-latest) | 4m 42s |
| Build (ubuntu-latest) | 9m 41s |
| Build (macos-latest) | 3m 7s |
| Build (ubuntu-24.04-arm) | 3m 32s |
| Unit Tests (windows-latest) | 3m 34s |
| Integration Tests (windows-latest, bash) | 8m 38s |
| Integration Tests (windows-latest, powershell) | 7m 14s |

**Total CI time: ~30 min → ~19 min (37% faster)**

## Test plan
- [x] Verify Windows build job completes and uploads artifacts
- [x] Verify unit test job restores cache and runs without recompiling
- [x] Verify both integration test jobs download binary and run
- [x] Compare total CI time against baseline


marcoesters (commented):
The reduction in overall time is great, but I have concerns about the added complexity to the workflow due to the added special-casing on Windows. The current workflow will also make Windows tests wait for the slowest build, which currently seems to be Ubuntu.

Here is an alternative refactor:
* `ci.yaml` becomes a workflow that only builds one binary at a time. That way, you can parallelize the tests. Let's call it `build-test.yaml`.
* `release.yaml` becomes the actual `ci.yaml` - you'd just have to add an if-condition to the conda upload that it's not triggered for merge groups.
* The new `ci.yaml` contains the matrix and calls `build-test.yaml` with the runner OS as the input.

We could also reduce the PowerShell or `git bash` scripts for Windows to the parts that are PowerShell/`git bash` exclusive to reduce test time. I would overall prefer that we test on PowerShell since that's the native Windows shell.

PR #189, opened 2026-05-08
fix: Handle broken pipe gracefully when output is piped

mattkram:
## Summary

Reset SIGPIPE to default behavior on Unix so the process terminates cleanly when output is piped to commands like `head` or `grep -q`.

Rust ignores SIGPIPE by default, which caused panics with "Broken pipe (os error 32)" when the receiving end of a pipe closed early.

## Test plan

- [ ] Run `ana tool list | head -1` - should exit cleanly with code 0
- [ ] Run `ana --help | head -1` - should exit cleanly with code 0
- [ ] Run `ana tool list | grep pixi` - should work without panic

Jira: [CLI-527](https://anaconda.atlassian.net/browse/CLI-527)

[CLI-527]: https://anaconda.atlassian.net/browse/CLI-527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -25,8 +25,23 @@ pub const VERSION: &str = env!("PKG_VERSION");
 #[cfg(feature = "feedback")]
 pub const FEEDBACK_BASE_URL: &str = "https://docs.google.com/forms/d/e/1FAIpQLSeGd9p7pQSHvjIc6RNShjTQCGmM-5_3xkPNpNfYk102-HZB8Q/viewform";
 
+/// Reset SIGPIPE to default behavior so the process terminates cleanly when
+/// output is piped to commands like `head` or `grep -q`. Rust ignores SIGPIPE
+/// by default, which causes panics on broken pipe errors.
+#[cfg(unix)]
+fn reset_sigpipe() {
+    unsafe {
+        libc::signal(libc::SIGPIPE, libc::SIG_DFL);
+    }
+}
+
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -25,8 +25,23 @@ pub const VERSION: &str = env!("PKG_VERSION");
 #[cfg(feature = "feedback")]
 pub const FEEDBACK_BASE_URL: &str = "https://docs.google.com/forms/d/e/1FAIpQLSeGd9p7pQSHvjIc6RNShjTQCGmM-5_3xkPNpNfYk102-HZB8Q/viewform";
 
+/// Reset SIGPIPE to default behavior so the process terminates cleanly when
+/// output is piped to commands like `head` or `grep -q`. Rust ignores SIGPIPE
+/// by default, which causes panics on broken pipe errors.
+#[cfg(unix)]
+fn reset_sigpipe() {
+    unsafe {
```
zfralish, line 33: Image

zfralish (approved):

src/main.rs
mattkram: Yeah, I saw this too. This is apparently a very common use of unsafe in Rust ecosystem. There are other ways, but require a dependency.

PR #191, opened 2026-05-08
fix: Validate API key against server before storing

mattkram:
## Summary

When using `ana login --api-key`, validate the key by making a request to `/api/account` before saving it to the keyring. This prevents storing invalid or cross-environment tokens that would later show "Logged in as unknown".

Before this fix, any valid JWT format would be accepted and stored, even if it was from a different environment or otherwise invalid for the target server.

## Test plan

- [ ] Run `ana login --api-key` with a valid API key - should succeed and show user email
- [ ] Run `ana login --api-key` with an invalid JWT - should show "API key rejected by server (4xx)"
- [ ] Run `ana login --api-key` with a key from a different environment - should show rejection error
- [ ] Pipe an invalid key: `echo "invalid.jwt.token" | ana login --force` - should show rejection error

Jira: [CLI-549](https://anaconda.atlassian.net/browse/CLI-549)

[CLI-549]: https://anaconda.atlassian.net/browse/CLI-549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/auth/actions.rs
+++ b/src/auth/actions.rs
@@ -180,30 +180,34 @@ fn prompt_api_key_hidden() -> Result<String, AuthError> {
 
 /// Validate an API key by making a request to the account endpoint.
 /// Returns the user's email if valid, or an error if invalid.
-async fn validate_api_key(ctx: &CommandContext, api_key: &str) -> Result<String, AuthError> {
+async fn validate_api_key(ctx: &CommandContext, api_key: &str) -> miette::Result<String> {
     let client = ctx.unauthenticated_client(REQUEST_TIMEOUT);
 
     let response = client
         .get("/api/account")
         .header("Authorization", format!("Bearer {}", api_key))
         .send()
-        .await?;
+        .await
+        .map_err(|e| miette::miette!("Failed to validate API key: {}", e))?;
 
     if !response.status().is_success() {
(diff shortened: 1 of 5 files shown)


src/auth/actions.rs
```diff
@@ -178,13 +178,44 @@ fn prompt_api_key_hidden() -> Result<String, AuthError> {
     Ok(api_key.trim().to_string())
 }
 
+/// Validate an API key by making a request to the account endpoint.
+/// Returns the user's email if valid, or an error if invalid.
+async fn validate_api_key(ctx: &CommandContext, api_key: &str) -> Result<String, AuthError> {
+    let client = ctx.unauthenticated_client(REQUEST_TIMEOUT);
+    let url = format!("{}/api/account", ctx.config.base_url());
```
zfralish: You shouldn't need to inject the base url here, I think you can just do relative, that client is built with the base url in it. 

Might be wrong though.

zfralish (approved):

src/auth/actions.rs
mattkram: Led to a bit of a sprawl of changes, but they look good. Mind double-checking my work here?

PR #193, opened 2026-05-08
feat: Sign and notarize macOS binary

marcoesters:
## Summary

Sign and notarize (for production builds) macOS binaries. Authentication is performed using environment secrets, which are restricted to releases for `prod` credentials. [Runtime hardening](https://developer.apple.com/documentation/security/hardened-runtime) is enabled for the binary.

This required me to remove the dependency `build-release` on the integration test task. `Cargo` would recompile the binary and overwrite the signature. This is also tripping up #182.

For dev builds, I created a self-signed application certificate and stored the data in Vault. I committed the script I used for posterity.

## Important notes

* Notarization cannot be tested with self-signed certificates. We will only know if it worked when creating a release.
* Running the binary on macOS the first time requires an internet connection because the notarization ticket cannot be stapled to a binary.
* The developer IDs and certificate fingerprints are not technically secrets. They are just stored in Vault for convenience.
* `conda` packages that are uploaded to the `dev` label will have the self-signed certificates. I can keep them unsigned if desired.

## AI Disclosure

Claude Code was used to review my code, write integration tests, and help debugging.

Jira: [CLI-604](https://anaconda.atlassian.net/browse/CLI-604)

[CLI-604]: https://anaconda.atlassian.net/browse/CLI-604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/sign-notarize-macos/action.yaml
+++ b/.github/actions/sign-notarize-macos/action.yaml
@@ -48,7 +48,7 @@ runs:
             echo "::error::Unrecognized vault mount point `${VAULT_MOUNT}`"
             exit 1
         fi
-        if ! [[ -f "${BINARY_PATH}" ]]; then
+        if [[ ! -f "${BINARY_PATH}" ]]; then
             echo "::error::${BINARY_PATH} does not exist."
             exit 1
         fi
--- a/.github/actions/sign-notarize-macos/action.yaml
+++ b/.github/actions/sign-notarize-macos/action.yaml
@@ -34,9 +34,6 @@ outputs:
   fingerprint:
     description: The fingerprint of the public certificate. Used for testing.
     value: ${{ steps.application-certificate.outputs.fingerprint }}
-  notarized:
-    description: Whether the application was notarized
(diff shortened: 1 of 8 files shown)


.github/actions/sign-notarize-macos/action.yaml
```diff
@@ -0,0 +1,180 @@
...
+
+runs:
+  using: composite
+  steps:
+    - name: Validate input
+      env:
+        BINARY_PATH: ${{ inputs.binary-path }}
+        VAULT_MOUNT: ${{ inputs.vault-mount }}
+      run: |
+        # Validate inputs
```
marcoesters, line 46: These initial comments are a little trick to make the log look nice. In composite actions, the `name` is ignored and instead the first line of the step is displayed:

```
> Run # Sign binary
```

mattkram (approved):
Wow, this is super cool. Great work!

PR #202, opened 2026-05-11
fix: Remove build-release dependency from conda package task

marcoesters:
## Summary

Remove the `build-release` prerequisite from the `build-conda` task. The macOS binaries that were uploaded to [v0.3.0](https://github.com/anaconda/ana-cli/releases/tag/v0.3.0) are not signed and notarized despite passing all integration tests. This is because before uploaded, the `conda` package build task recompiled `ana` and overwrites the signed binary. For local testing, that binary will have to be compiled now before running the task.

I also found that the release was not set as `latest` automatically, probably because prior releases had been manually set, so I re-added that flag to the `gh` call.

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -37,7 +37,7 @@ jobs:
       - name: Set up ana with pixi
         uses: anaconda/github-actions/setup-ana@3d7b08cd24c1c5fc8a92a20caa9564bda187d7c1 # v0.2.1
         with:
-          ana-version: v0.2.0
+          ana-version: v0.1.2
           tools: pixi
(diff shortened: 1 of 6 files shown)


.github/workflows/pre-commit.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
       - name: Set up ana with pixi
         uses: anaconda/github-actions/setup-ana@3d7b08cd24c1c5fc8a92a20caa9564bda187d7c1 # v0.2.1
         with:
-          ana-version: v0.2.0
+          ana-version: v0.1.2
```
mattkram, line 40: sorry 🙈

mattkram (approved):
Looks great.

PR #205, opened 2026-05-11
fix: Expand shim approach to `conda` wrapper

marcoesters:
## Summary

Expand the shimming approach on Windows to the `conda` wrapper. Similar to the symlink approach, the shim points to the `ana` binary. A consequence is that the `is_conda_invocation` function would see `ana.exe` as the calling binary, so it would not recognize that `ana` is being called as a wrapper. To circumvent this, we use an environment variable to indicate which wrapper is being called. This environment variable is short-lived and is removed before calling the `conda` file.

This PR also makes other small fixes to the `conda` implementation:

* Use the newly available packages of `conda-spawn` and `conda-self` on `main`. This removes `conda-forge` as a dependency.
* Update tests to work on Windows and for local testing.

Jira: [CLI-402](https://anaconda.atlassian.net/browse/CLI-402)

## AI disclosure

Claude Code was used to review the code, help debugging, and to write unit tests.

[CLI-402]: https://anaconda.atlassian.net/browse/CLI-402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/tools/install.rs
```diff
@@ -214,6 +217,41 @@ fn create_bin_symlink(bin_dir: &Path, prefix: &Path, binary: &Path) -> miette::R
     Ok(())
 }
 
+#[cfg(unix)]
+/// Create a symlink that points to the ana binary itself.
+///
+/// This is used for tools that ana wraps (like conda), where ana detects
```
mattkram, line 223: As we continue to learn and evaluate e.g. cx, I think we will want to be extremely careful about how we wrap conda.

We technical wrap it with ana here, but that is someone for convenience since it is a binary we have available. But I do want to make sure we don't introduce layers of complexity that will be confusing to end users, or potentially introduce liability and instability to conda.

For now, I think it makes sense as a technical exploration, but before we ship this to users we will want a much more comprehensive technical review.

(I'm saying this more for myself, not as any particular criticism of the approach here).

mattkram (approved):

src/tools/install.rs
marcoesters: For sure, that's why I created this as a PR against your draft branch. It's what is currently needed, but the added complexity is not insignificant.

PR #209, opened 2026-05-12
chore: Gate feature behind unstable compile flag

mattkram:
## Summary
- Adds `unstable` Cargo feature flag to gate the wheels functionality
- When built without `--features unstable`, `ana feature enable wheels` returns "Unknown feature"
- When built with `--features unstable`, wheels feature works as before
- Gates related modules (`pip`, `uv`, `utils`) and `checkmark()` helper to avoid dead code warnings

## Test plan
- [x] `cargo test` passes (198 tests without unstable, 220 with)
- [x] `cargo clippy` passes with no unused warnings
- [x] `cargo run -- feature enable wheels` returns "Unknown feature: wheels"
- [x] `cargo run --features unstable -- feature enable wheels` prompts for experimental feature

### DIFF
--- a/src/ui/status.rs
+++ b/src/ui/status.rs
@@ -73,6 +73,7 @@ pub fn highlight(text: &str) -> String {
 }
 
 /// Return a green checkmark.
+#[cfg(feature = "unstable")]
 pub fn checkmark() -> String {
     UiColor::Green.apply_to("✓").to_string()
 }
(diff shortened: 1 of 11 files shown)


src/ui/status.rs
```diff
@@ -73,6 +73,7 @@ pub fn highlight(text: &str) -> String {
 }
 
 /// Return a green checkmark.
+#[cfg(feature = "unstable")]
```
zfralish, line 76: This looks like a debug thing, did you mean to leave this here?

zfralish (approved):

zfralish (approved):

src/ui/status.rs
mattkram: Yes. It's just because we don't use this function anywhere else right now. We should put it into the ui module but for now this is okay.

PR #211, opened 2026-05-12
feat: Add public telemetry endpoint for unauthenticated users

zfralish:
## Summary
- Add separate telemetry endpoint for unauthenticated users (`public.telemetry.anaconda.com`)
- Authenticated users continue to use `metrics.aa.anaconda.com` with their API key
- Only set auth token when an API key exists
- Both endpoints are configurable via environment variables (`ANA_METRICS_ENDPOINT` and `ANA_METRICS_PUBLIC_ENDPOINT`)

## Test plan
- [ ] Verify telemetry is sent to public endpoint when not logged in
- [ ] Verify telemetry is sent to authenticated endpoint with auth token when logged in
- [ ] Verify `ANA_METRICS_PUBLIC_ENDPOINT` env var overrides the default public endpoint

Jira: [CLI-441](https://anaconda.atlassian.net/browse/CLI-441)

[CLI-441]: https://anaconda.atlassian.net/browse/CLI-441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/telemetry/otel.rs
+++ b/src/telemetry/otel.rs
@@ -61,10 +61,19 @@ pub fn setup() {
 fn try_setup() -> Result<(), Box<dyn std::error::Error>> {
     let app_config = crate::config::Config::load();
 
-    let mut otel_config = Configuration::new(Some(&app_config.metrics_endpoint), None)?;
-
     let api_key = crate::auth::get_api_key(&app_config).ok().flatten();
-    otel_config.set_auth_token(api_key);
+
+    let endpoint = if api_key.is_some() {
+        &app_config.metrics_endpoint
+    } else {
+        &app_config.metrics_public_endpoint
+    };
+
+    let mut otel_config = Configuration::new(Some(endpoint), None)?;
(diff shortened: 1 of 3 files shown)


src/telemetry/otel.rs
```diff
@@ -61,10 +61,19 @@ pub fn setup() {
...
+        &app_config.metrics_endpoint
+    } else {
+        &app_config.metrics_public_endpoint
+    };
+
+    let mut otel_config = Configuration::new(Some(endpoint), None)?;
+
+    if let Some(key) = api_key {
+        otel_config.set_auth_token(Some(key));
+    }
```
mattkram, line 76: We probably could consolidate these two conditionals, but it's probably not worth over-optimizing.

mattkram (approved):
Looks good

PR #220, opened 2026-05-12
ci: Skip running integration test in Git Bash

marcoesters:
## Summary

Windows currently runs the full set of integration tests in both Git Bash and PowerShell. As pointed out in #182, this takes a long time to finish. #182 introduced parallelization of integration test, but it comes with increased code complexity.

Initially, I added the PowerShell tests to test the PowerShell installation script and kept Git Bash because I erroneously believed that the SH installation script test didn't run on PowerShell. Looking at the logs though, that appears to be false: they run and succeed inside PowerShell. So, I think there is little gain in running the full set of integration tests in both shells.

Instead, run on workflows on default shells (PowerShell on Windows and `bash` on Linux) unless `bash` is specifically required.

Supersedes #182

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -43,7 +43,7 @@ jobs:
       run:
         # Use native shells unless bash is necessary.
         # Prefer PowerShell 5 on Windows since this is the default version.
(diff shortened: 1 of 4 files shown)


mattkram (approved):
This is a lot cleaner! This should improve test speeds a good amount 😄

PR #207, opened 2026-05-11
feat: Sign windows binaries

marcoesters:
## Summary

Sign Windows binaries using AzureSignTool. This requires a change to the way Windows binaries are compiled: since the shim is a binary executable, compiling `ana` would lead to unsigned shim binaries. Instead, the shim is compiled and signed separately and then directly included into the resulting `ana` binary.

The integration tests do not perform any thumbprint verification for now because I do not have access to the certificates directly. If I can get those before this PR is merged, I will revert https://github.com/anaconda/anaconda-cli/pull/207/commits/4527d460fbaeeaba82274aa5b04ac9c422e3b725

I also cleaned up some things in the macOS signing tests.

Jira: [CLI-367](https://anaconda.atlassian.net/browse/CLI-367)

## AI disclosure

Claude Code was used to review code and make the changes needed to `build.rs`.

[CLI-367]: https://anaconda.atlassian.net/browse/CLI-367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #221, opened 2026-05-13
chore: Add fingerprint test for Windows certificates

marcoesters:
## Summary

Verify the signature thumbprint in Windows integration tests. Windows signing was introduced in #207, but lacked fingerprint verification the macOS signing tests had. Now that the thumbprints are available in the Vault, we can test the fingerprint the same way.


mattkram (approved):
Nice

PR #224, opened 2026-05-13
fix: Skip download when updating to same version

mattkram:
## Summary
- When running `ana self update <version>` where `<version>` matches the current version, show "UP TO DATE" instead of downloading and reinstalling unnecessarily
- Add `--force` flag to bypass the same-version check for cases where reinstallation is desired (e.g., repairing a corrupted installation)

## Test plan
- [x] `cargo test` passes
- [x] `cargo clippy` passes
- [ ] Run `ana self update v<current>` and verify it shows "UP TO DATE" without downloading
- [ ] Run `ana self update v<current> --force` and verify it proceeds with the download
- [ ] Run `ana self update --help` and verify `--force` flag is documented

Jira: [CLI-626](https://anaconda.atlassian.net/browse/CLI-626)

[CLI-626]: https://anaconda.atlassian.net/browse/CLI-626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/test_self_update.py
+++ b/tests/test_self_update.py
@@ -54,6 +54,13 @@ class TestSelfUpdateSameVersion:
             or "Already on version" not in result.stderr
         )
 
+    def test_update_check_still_works(self, run_ana: AnaRunner) -> None:
+        """The --check flag should still work as before."""
+        result = run_ana("self", "update", "--check")
+        assert result.returncode == 0
+        # Should show either update available or up to date
+        assert "UPDATE" in result.stderr or "UP TO DATE" in result.stderr
+
 
 class TestSelfUpdateCli:
     """Tests for self update CLI interface (no network required)."""
@@ -87,10 +94,3 @@ class TestSelfUpdateDifferentVersion:
         result = run_ana("self", "update", "v999.999.999")
         assert result.returncode == 0  # Command runs but shows error
         assert "not found" in result.stderr.lower()
-
-    def test_update_check_still_works(self, run_ana: AnaRunner) -> None:
-        """The --check flag should still work as before."""
-        result = run_ana("self", "update", "--check")
-        assert result.returncode == 0
-        # Should show either update available or up to date
-        assert "UPDATE" in result.stderr or "UP TO DATE" in result.stderr
--- a/tests/test_self_update.py
+++ b/tests/test_self_update.py
@@ -2,19 +2,19 @@
 
 from __future__ import annotations
 
+import pytest
 from helpers import AnaRunner
 
 
 class TestSelfUpdateSameVersion:
     """Tests for 'ana self update <version>' when already on that version."""
 
+    @pytest.mark.parametrize("prefix", ["v", ""])
     def test_update_to_same_version_shows_up_to_date(
-        self, run_ana: AnaRunner, ana_version: str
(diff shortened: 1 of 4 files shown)


tests/test_self_update.py
```diff
@@ -0,0 +1,102 @@
+"""Integration tests for ana self update same-version handling."""
+
+from __future__ import annotations
+
+from helpers import AnaRunner
+
+
+def get_version(run_ana: AnaRunner) -> str:
```
marcoesters: Could be a session-scoped or at least a class-scoped fixture.

tests/test_self_update.py
```diff
@@ -0,0 +1,102 @@
...
+    """Get the current ana version."""
+    result = run_ana("--version")
+    assert result.returncode == 0
+    return result.stdout.strip()
+
+
+class TestSelfUpdateSameVersion:
+    """Tests for 'ana self update <version>' when already on that version (CLI-626)."""
```
marcoesters:
```suggestion
    """Tests for 'ana self update <version>' when already on that version."""
```

tests/test_self_update.py
```diff
@@ -0,0 +1,102 @@
...
+
+        # Should show "UP TO DATE" status, not "UPDATED"
+        assert "UP TO DATE" in result.stderr
+        assert "UPDATED" not in result.stderr
+        # Should show current version
+        assert current_version in result.stderr
+        # Should NOT show download progress
+        assert "Downloading" not in result.stderr
+
+    def test_update_to_same_version_without_v_prefix(self, run_ana: AnaRunner) -> None:
```
marcoesters: Instead of two separate tests, you could parametrize it along the version prefix.

tests/test_self_update.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def test_update_to_nonexistent_version_shows_error(
+        self, run_ana: AnaRunner
+    ) -> None:
+        """Updating to a non-existent version should show an error."""
+        result = run_ana("self", "update", "v999.999.999")
+        assert result.returncode == 0  # Command runs but shows error
+        assert "not found" in result.stderr.lower()
+
+    def test_update_check_still_works(self, run_ana: AnaRunner) -> None:
```
marcoesters: This seems to belong more in the `TestSelfUpdateSameVersion` class than here.

marcoesters (commented):
The code looks good, I just have a few suggestions to streamline the tests.

tests/test_self_update.py
mattkram: I made it a fixture. We can't make it a session-scoped because `run_ana` is function-scoped for env var isolation. Feels fine to do this since I'm not anticipating performance concerns vs. complexity of a "runner-factory" pattern.

tests/test_self_update.py
mattkram: Good idea, done.

tests/test_self_update.py
mattkram: Good idea, done.

marcoesters (approved):

PR #225, opened 2026-05-13
fix: Raise RLIMIT_NOFILE to prevent 'too many open files' on macOS

mattkram:
## Summary
Raises the soft file descriptor limit from macOS's default of 256 to 10,240 (or the hard limit, whichever is lower) before package installation begins. This prevents "too many open files" errors during rattler installations.

Uses the `rlimit` crate, which is the same approach used by pixi.

Fixes #223

## Test plan
- [ ] Run `ana tool install pixi` on macOS and verify no fd exhaustion errors
- [ ] Verify the limit is raised by running with `RUST_LOG=debug` and checking for the `RLIMIT_NOFILE raised` log message

Jira: [CLI-624](https://anaconda.atlassian.net/browse/CLI-624)

[CLI-624]: https://anaconda.atlassian.net/browse/CLI-624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -38,9 +38,32 @@ fn reset_sigpipe() {
 #[cfg(not(unix))]
 fn reset_sigpipe() {}
 
+/// Attempt to raise RLIMIT_NOFILE to a value suitable for rattler installations.
+/// On macOS, respects kern.maxfilesperproc (the real hard ceiling).
+/// Logs a warning if the limit could not be raised; never fatal.
+#[cfg(unix)]
+pub fn raise_open_file_limit(target: u64) -> u64 {
+    match rlimit::increase_nofile_limit(target) {
+        Ok(n) => {
+            tracing::debug!(limit = n, "RLIMIT_NOFILE raised");
+            n
+        }
+        Err(e) => {
+            tracing::warn!(error = %e, target, "Failed to raise RLIMIT_NOFILE");
+            0
+        }
+    }
+}
+
+#[cfg(not(unix))]
+pub fn raise_open_file_limit(_target: u64) -> u64 {
+    0
+}
+
 #[tokio::main]
 async fn main() {
     reset_sigpipe();
+    raise_open_file_limit(10240);
 
     let config = config::Config::load();
     let _diagnostics_guard = diagnostics::init(&config);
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -38,9 +38,24 @@ fn reset_sigpipe() {
...
+    }
+}
+
+#[cfg(not(unix))]
+fn raise_open_file_limit(_target: u64) {}
+
 #[tokio::main]
 async fn main() {
     reset_sigpipe();
+    raise_open_file_limit(10240);
```
marcoesters: That limit seems a lot higher than what prefix uses: https://github.com/prefix-dev/pixi/blob/e640477b0f706982925d87105dc862d24a888fc2/crates/pixi_utils/src/rlimit.rs

Why is our limit higher?

src/main.rs
```diff
@@ -38,9 +38,24 @@ fn reset_sigpipe() {
...
+#[cfg(unix)]
+fn raise_open_file_limit(target: u64) {
+    match rlimit::increase_nofile_limit(target) {
+        Ok(n) => tracing::debug!(limit = n, "RLIMIT_NOFILE raised"),
+        Err(e) => tracing::warn!(error = %e, target, "Failed to raise RLIMIT_NOFILE"),
+    }
+}
+
+#[cfg(not(unix))]
+fn raise_open_file_limit(_target: u64) {}
```
marcoesters: This is the second "setup" function that we use that do not apply for Windows. Instead of creating empty placeholders, should we just put these function into a `#[cfg(not(unix))]` block at the beginning of `main`?

src/main.rs
mattkram: This was a common macOS maximum. But I agree aligning on the same approach pixi has chosen is a better approach.

src/main.rs
mattkram: I think that's a good idea. Maybe I'll refactor to a `setup.rs` module.

mattkram: I did refactor, but didn't create a new module. We can continue refining more as we review the code structure given recent learnings.

marcoesters (approved):

PR #229, opened 2026-05-14
fix: Support file input for -j flag to fix PowerShell compatibility

zfralish:
## Summary
- PowerShell handles single-quoted JSON strings differently from bash, causing `ana api fetch -j '{"foo":"bar"}'` to fail with "Invalid JSON" on Windows
- Added support for reading JSON from files using `@file.json` syntax (following curl's convention)
- Added support for reading JSON from stdin using `-` or `@-`
- Existing direct JSON string input continues to work unchanged

## Test plan
- [x] `cargo test` passes (202 tests)
- [x] `cargo clippy` passes
- [ ] On Windows PowerShell: `echo '{"foo":"bar"}' > body.json && ana api fetch /api/test --method POST -j @body.json`
- [ ] On bash/zsh: `ana api fetch /api/test --method POST -j '{"foo":"bar"}'` still works

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1013,7 +1013,7 @@ enum ApiCommands {
         #[arg(short = 'd', long, conflicts_with = "json")]
         data: Option<String>,
 
-        /// JSON request body
+        /// JSON request body (use @file.json to read from file, or - for stdin)
         #[arg(short = 'j', long, conflicts_with = "data")]
         json: Option<String>,
     },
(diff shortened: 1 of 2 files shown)


marcoesters (commented):
I think this needs integration tests to show that this actually works with command line input. When I run the `api fetch` command with this branch, I still get an invalid JSON error:

```
& .\target\release\ana.exe api fetch -j '{"foo":"bar"}' https://anaconda.com/api
Error:   × Invalid JSON
  ╰─▶ key must be a string at line 1 column 2
```

mattkram (commented):
I'm not convinced this is something we should change. Powershell requires double quotes and escaping when constructing the prompt, but I'm not sure that we want to do anything special here.

PR #230, opened 2026-05-14
fix: Warn when anaconda-cli is outdated in ana mcp

zfralish:
## Summary
After `ana self update`, managed tools like anaconda-cli are not automatically updated. This caused `ana mcp setup` to fail with cryptic errors ("No such command 'mcp'") when the installed anaconda-cli lacked the mcp subcommand.

This PR adds version checking to `ana mcp` that compares the installed anaconda-cli version against the embedded lockfile version. When outdated, it shows a clear warning:

```
! anaconda-cli is outdated (installed: 0.8.1). Run: ana tool install anaconda-cli
```

## Test plan
- [ ] Install an older version of anaconda-cli, then update ana to a version with a newer embedded lockfile
- [ ] Run `ana mcp setup` and verify the warning appears
- [ ] Run `ana tool install anaconda-cli` and verify `ana mcp setup` works without warning
- [ ] Verify `cargo test` passes (201 tests)
- [ ] Verify `cargo clippy` has no new warnings

### DIFF
--- a/src/mcp/run.rs
+++ b/src/mcp/run.rs
@@ -10,13 +10,14 @@ pub async fn run(ctx: &mut CommandContext, args: &[String]) -> miette::Result<()
         status::info("Installing anaconda-cli...");
         tools::install::install_tool(ctx, "anaconda-cli").await?;
         status::blank_line();
-    } else if let Some(installed_version) = tools::tools::package_name("anaconda-cli")
-        .and_then(|pkg| tools::install::check_tool_needs_update("anaconda-cli", pkg))
+    } else if let Some((installed_version, min_required)) =
+        tools::install::check_tool_incompatible("anaconda-cli")
     {
         status::warn(&format!(
-            "anaconda-cli is outdated (installed: {}). Run: ana tool install anaconda-cli",
-            installed_version
+            "anaconda-cli {} is incompatible with this version of ana (requires >= {})",
+            installed_version, min_required
         ));
+        eprintln!("  Run: ana tool install anaconda-cli");
         status::blank_line();
     }
 
--- a/src/mcp/run.rs
+++ b/src/mcp/run.rs
@@ -10,8 +10,8 @@ pub async fn run(ctx: &mut CommandContext, args: &[String]) -> miette::Result<()
         status::info("Installing anaconda-cli...");
         tools::install::install_tool(ctx, "anaconda-cli").await?;
         status::blank_line();
-    } else if let Some(installed_version) =
-        tools::install::check_tool_needs_update("anaconda-cli", "anaconda-cli-base")
+    } else if let Some(installed_version) = tools::tools::package_name("anaconda-cli")
+        .and_then(|pkg| tools::install::check_tool_needs_update("anaconda-cli", pkg))
     {
         status::warn(&format!(
             "anaconda-cli is outdated (installed: {}). Run: ana tool install anaconda-cli",
--- a/src/mcp/run.rs
+++ b/src/mcp/run.rs
@@ -10,6 +10,14 @@ pub async fn run(ctx: &mut CommandContext, args: &[String]) -> miette::Result<()
         status::info("Installing anaconda-cli...");
         tools::install::install_tool(ctx, "anaconda-cli").await?;
         status::blank_line();
+    } else if let Some(installed_version) =
+        tools::install::check_tool_needs_update("anaconda-cli", "anaconda-cli-base")
+    {
+        status::warn(&format!(
+            "anaconda-cli is outdated (installed: {}). Run: ana tool install anaconda-cli",
+            installed_version
+        ));
+        status::blank_line();
     }
 
     let mut mcp_args = vec!["mcp".to_string()];
--- a/src/tools/install.rs
+++ b/src/tools/install.rs
@@ -16,6 +16,21 @@ use super::{pixi_config, tools};
 use crate::context::CommandContext;
 use crate::paths;
 
+/// Check all installed tools and return the names of those that need updating.
+///
+/// Returns a list of tool names that are installed but at a different version
+/// than what's in the embedded lockfile.
+pub fn check_all_tools_need_update() -> Vec<&'static str> {
+    tools::all_tools()
+        .into_iter()
+        .filter(|tool_name| {
+            tools::package_name(tool_name)
+                .and_then(|pkg| check_tool_needs_update(tool_name, pkg))
+                .is_some()
+        })
+        .collect()
+}
+
 /// Check if a tool needs updating by comparing installed vs embedded lockfile versions.
 ///
 /// Returns `Some(installed_version)` if the tool is installed but at a different version
--- a/src/tools/install.rs
+++ b/src/tools/install.rs
@@ -16,6 +16,53 @@ use super::{pixi_config, tools};
 use crate::context::CommandContext;
 use crate::paths;
 
+/// Check if a tool needs updating by comparing installed vs embedded lockfile versions.
+///
+/// Returns `Some(installed_version)` if the tool is installed but at a different version
+/// than what's in the embedded lockfile. Returns `None` if the tool is not installed
+/// or is already at the correct version.
+pub fn check_tool_needs_update(tool_name: &str, package_name: &str) -> Option<String> {
+    let prefix = paths::tool_prefix(tool_name);
+    if !prefix.exists() {
+        return None;
+    }
+
+    // Get installed version from prefix
+    let installed = PrefixRecord::collect_from_prefix::<PrefixRecord>(&prefix).ok()?;
+    let installed_version = installed
+        .iter()
+        .find(|r| r.repodata_record.package_record.name.as_normalized() == package_name)
+        .map(|r| r.repodata_record.package_record.version.to_string())?;
+
+    // Get expected version from embedded lockfile
+    let lock_content = tools::content(tool_name)?;
+    let expected_version = get_package_version_from_lockfile(&lock_content, package_name)?;
+
+    if installed_version != expected_version {
+        Some(installed_version)
+    } else {
+        None
+    }
+}
+
+/// Extract a package version from a lockfile string for the current platform.
+fn get_package_version_from_lockfile(lock_content: &str, package_name: &str) -> Option<String> {
+    let lock_file = LockFile::from_str_with_base_directory(lock_content, None).ok()?;
+    let env = lock_file.default_environment()?;
+    let current_platform = Platform::current();
+    let records_by_platform = env.conda_repodata_records_by_platform().ok()?;
+
+    let records = records_by_platform
+        .into_iter()
+        .find(|(p, _)| p.subdir() == current_platform)
+        .map(|(_, records)| records)?;
+
+    records
+        .iter()
+        .find(|r| r.package_record.name.as_normalized() == package_name)
+        .map(|r| r.package_record.version.to_string())
+}
+
 /// Global progress bar for installation feedback.
 static MULTI_PROGRESS: std::sync::LazyLock<MultiProgress> = std::sync::LazyLock::new(|| {
     let mp = MultiProgress::new();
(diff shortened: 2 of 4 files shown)


src/tools/install.rs
```diff
@@ -16,6 +16,53 @@ use super::{pixi_config, tools};
...
+    let installed_version = installed
+        .iter()
+        .find(|r| r.repodata_record.package_record.name.as_normalized() == package_name)
+        .map(|r| r.repodata_record.package_record.version.to_string())?;
+
+    // Get expected version from embedded lockfile
+    let lock_content = tools::content(tool_name)?;
+    let expected_version = get_package_version_from_lockfile(&lock_content, package_name)?;
+
+    if installed_version != expected_version {
```
marcoesters: Is strict equality necessary here? What if the installed package is newer than what's in the lockfile?

src/mcp/run.rs
```diff
@@ -10,6 +10,14 @@ pub async fn run(ctx: &mut CommandContext, args: &[String]) -> miette::Result<()
         status::info("Installing anaconda-cli...");
         tools::install::install_tool(ctx, "anaconda-cli").await?;
         status::blank_line();
+    } else if let Some(installed_version) =
+        tools::install::check_tool_needs_update("anaconda-cli", "anaconda-cli-base")
+    {
+        status::warn(&format!(
+            "anaconda-cli is outdated (installed: {}). Run: ana tool install anaconda-cli",
+            installed_version
+        ));
```
marcoesters, line 19: I think we will run into this problem often where `ana` ships a capability that requires newer versions of tools. Warnings are good for interactive use even though it might be annoying to users that they have to update tools manually instead of updates "just working". I luckily don't see the divergence break non-interactive workflows though.

Ideally, we'd want to check that prior to self-updating. So, I think the real capability we need to implement here is a compatibility check during the self-update procedure. I'm happy to have an issue created for now and apply this band-aid for the immediate fix.

src/tools/install.rs
zfralish: Do you think `installed_version >= expected_version` be better?

src/mcp/run.rs
zfralish: If we prefer the approach of just warning and having users manually resolve this with guidance, I am fine with that. I was just thinking that with regard to tool we directly manage, the capability to automatically update them would be something the user would expect, but will defer to you or @mattkram to decide what is a better appraoch.

zfralish: But for now I will move this logic to update as well, as I don't think it will cause issues and accomplishes the capability you are suggesting.

src/tools/install.rs
marcoesters: I think so. If a user has a newer version, the warning that the tool is outdated is not correct. The downside of `>=` is that we assume some degree of backwards compatibility. For `anaconda-cli`, that's probably a reasonable assumption.

zfralish: Cool, will push that up.

src/mcp/run.rs
mattkram: I've been thinking that, what we want is a conception of "tool update during `ana self update`". Since all the current tools are intended to be carefully versioned and managed via ana and its lockfiles, I think we should update the embedded tools that are already installed when we update ana.

This will get more complex though when we have tools that manage their own updates. We need to carefully think through that problem, because we are not trying to build a general-purpose package manager, just something that uses similar technologies in a very scoped/restricted way.

For the near-term, I recommend we:
* Add a flag on the tool specs to indicate "fully-managed" (can also be a TODO comment, since all current tool specs are designed to be fully managed)
* Determine if this should be configurable or disable-able. 
* For the specific bug we should just make sure things like outerbounds and anaconda-cli get rolled forward if the lockfile changes. I think rattler does this for us, but we should validate.

I consider these current specs as "carefully crafted locks of current versions of these tools you can install w/ ana". We need to support update/sync to whatever version is embedded in the ana version you are using, but we don't need/want to extend that further at this point without further design.

marcoesters: > I think we should update the embedded tools that are already installed when we update ana.

I'm not sure about that - updating a tool without the user's consent isn't the greatest UX. I'd rather this be opt-in (like `ana self update --update-tools`) unless updating `ana` would break a tool.

mattkram: I'm okay with that. Start with a notification, and explicit way to update. Potentially explore an "auto-update" opt-in configuration feature down the road.

src/tools/install.rs
```diff
@@ -16,6 +16,68 @@ use super::{pixi_config, tools};
...
+        })
+        .collect()
+}
+
+/// Check if a tool needs updating by comparing installed vs embedded lockfile versions.
+///
+/// Returns `Some(installed_version)` if the tool is installed but at a different version
+/// than what's in the embedded lockfile. Returns `None` if the tool is not installed
+/// or is already at the correct version.
+pub fn check_tool_needs_update(tool_name: &str, package_name: &str) -> Option<String> {
```
marcoesters: The premise here seems to be that the tool is always going to be a `conda` package. This is true for now, but I think a more future-proof way would be to only require `check_tool_needs_update` to require the tool name and detect the package name inside the function. That would keep the method to check for updates localized.

src/update.rs
```diff
@@ -411,6 +411,27 @@ fn print_update_success(current_version: &str, new_version: &str, elapsed: std::
...
+
+    let outdated = check_all_tools_need_update();
+    if outdated.is_empty() {
+        return;
+    }
+
+    let tool_list = outdated.join(", ");
+
+    status::warn(&format!(
+        "The following tools may also need updating: {}",
```
marcoesters: This unfortunately doesn't give the user much actionable feedback. The tools are not the latest, but there are different levels of urgency between not having the latest version and having an incompatible version. The PR was created to address the latter, not the former.

marcoesters (commented):
I'm not sure these changes go into the right direction. The PR was created because the `anaconda-cli` tool was incompatible with the latest version of `ana`, so updating `ana` would leave the user in a broken state. The current changes list all tools that are outdated, which is a big difference.

I do like the more generalized solution about update checks though and we can leverage that. I think the best way forward is to bake in some minimum requirements for `ana`, use your new functions to check for tool versions, and auto-update them (with a prompt to confirm?). That way, `ana update self` doesn't result in a broken state and doesn't end up with undesired tool updates.

src/tools/install.rs
zfralish: Okay, I have done what you suggested in the most recent commit

src/update.rs
zfralish: Okay updated this to give more actionable info

src/tools/tools.rs
```diff
@@ -5,36 +5,48 @@ use std::path::PathBuf;
...
+        package_name: "anaconda-cli-base",
         lockfile: include_str!("../../tool-specs/anaconda-cli/pixi.lock"),
         binaries: if cfg![unix] {
             &[&["bin", "anaconda"]]
         } else {
             &[&["Scripts", "anaconda"]]
         },
         experimental: None,
+        // anaconda-cli 0.8.1+ required for MCP support
+        min_compatible_version: Some("0.8.1"),
```
mattkram: I think we need to consider this more deeply. What we really are versioning here is like a metapackage. We have not actually encapsulated `anaconda-cli` as a metapackage (it actually used to be a repo with this name, but we renamed it to https://github.com/anaconda/anaconda-cli-metapackage and it hasn't been released). The current assumption is we will treat the lockfile here as the metapackage (at least currently), which doesn't have a version.

For now, what I think we should do is to leverage existing mechanism to just check the lockfile hash as an indicator of the version. If we find that is insufficient, we can pursue a metapackage that describes all of the dependencies and version minima. I believe rattler can do this natively already. If not, we should consider writing the lockfile hash somewhere in the installed environment for tracking, but I'd work to exhaust rattler's native capabilities first.

src/tools/install.rs
```diff
@@ -16,6 +16,97 @@ use super::{pixi_config, tools};
...
+/// Returns `Some(installed_version)` if the tool is installed but at a lower version
+/// than what's in the embedded lockfile. Returns `None` if the tool is not installed
+/// or is already at or above the lockfile version.
+#[allow(dead_code)] // Reserved for future --update-tools flag
+pub fn check_tool_needs_update(tool_name: &str) -> Option<String> {
+    let package_name = tools::package_name(tool_name)?;
+    let prefix = paths::tool_prefix(tool_name);
+    if !prefix.exists() {
+        return None;
+    }
```
marcoesters:
```suggestion
    let prefix = paths::tool_prefix(tool_name);
    if !prefix.exists() {
        return None;
    }
    
    let package_name = tools::package_name(tool_name)?;
```

If `prefix`doesn't exist, we would perform an unnecessary lookup.

src/tools/install.rs
```diff
@@ -16,6 +16,97 @@ use super::{pixi_config, tools};
...
+}
+
+/// Check if a tool is incompatible with this version of ana.
+///
+/// Returns `Some((installed_version, min_required))` if the tool is installed but
+/// below the minimum compatible version. Returns `None` if the tool is not installed,
+/// has no minimum version requirement, or is compatible.
+pub fn check_tool_incompatible(tool_name: &str) -> Option<(String, &'static str)> {
+    let min_required = tools::min_compatible_version(tool_name)?;
```
marcoesters:
```suggestion
    let min_required = tools::min_compatible_version(tool_name)?;
    if min_required.is_none() {
        None
    }
```

Always good to return early.

src/tools/install.rs
```diff
@@ -16,6 +16,97 @@ use super::{pixi_config, tools};
...
+    let package_name = tools::package_name(tool_name)?;
+    let prefix = paths::tool_prefix(tool_name);
+    if !prefix.exists() {
+        return None;
+    }
```
marcoesters:
```suggestion
    let prefix = paths::tool_prefix(tool_name);
    if !prefix.exists() {
        return None;
    }
    
    let package_name = tools::package_name(tool_name)?;
```

Same as above.

src/tools/install.rs
```diff
@@ -16,6 +16,97 @@ use super::{pixi_config, tools};
...
+        })
+        .collect()
+}
+
+/// Check if a tool needs updating by comparing installed vs embedded lockfile versions.
+///
+/// Returns `Some(installed_version)` if the tool is installed but at a lower version
+/// than what's in the embedded lockfile. Returns `None` if the tool is not installed
+/// or is already at or above the lockfile version.
+#[allow(dead_code)] // Reserved for future --update-tools flag
```
marcoesters: Which of that code is dead-code? I generally would prefer to not add interfaces that aren't used yet.

src/tools/tools.rs
marcoesters: I agree that there is the risk of a semantic overload. Would `min_compatible_package_version` be better? That would provide some semantic mapping that this refers to the sentinel package.

marcoesters (commented):
Mostly code optimization comments, but Matt has a point about being careful about what we're versioning.

src/tools/install.rs
zfralish: Okay, will remove. This was used prior to the refactor, in another comment you said you liked the approach so just left as is in case we wanted to use it in the future.

src/tools/tools.rs
mattkram: The problem there then is, version of which package? For example, the actual source of the reported issue here was "anaconda-mcp was not in the original lockfile spec". What we are actually doing is trying to understand that some collective set of packages needs to be updated on the system.

So in my mind, we basically have two alternatives:
* Create a metapackage, where we define these packages, with version constraints. Version this metapackage on its own and track that. This is more of a "you need to update" thing.
* Just use the embedded lockfile we have, which is explicitly tied to whatever version of `ana` you are running. And which implicitly has a hash associated with it that rattler uses to determine whether it needs to be updated or not. The behavior here is more like: synchronize my tool environment with whatever spec/lockfile my version of `ana` intents it to be.

Personally, since we are explicitly controlling every single version right now via the cross-platform lockfile, managed via Renovate updates, I think we should go with the second option. If we find later that we need to be a bit looser where a separately-versioned metapackage is more appropriate, we can always change it. This lets us keep the specific tools that `ana` needs to be tightly controlled, with minimal variability. This feels appropriate since the current general case is for wrapped legacy tools, and not direct user-facing tools (at least for anaconda-cli).

marcoesters: > The problem there then is, version of which package? 

The package given in `package_name`.

> The behavior here is more like: synchronize my tool environment with whatever spec/lockfile my version of ana intents it to be.

The consequence of that is that with each update to `anaconda-cli-base` or other dependencies, we need a new version of `ana`.

PR #234, opened 2026-05-15
fix: Cache user_id in keyring

zfralish:
## Summary
- After login, fetch `user_id` from `/api/auth/sessions/whoami` and store it in the keyring alongside the API key
- On subsequent command runs, load the cached `user_id` and add it as a telemetry attribute
- The `user_id` is cleared on logout (since the entire credential entry is removed)
- Backwards compatible with existing keyrings via `#[serde(default)]`

## Test plan
- [ ] `ana login` → verify user_id is fetched and stored (check keyring file)
- [ ] Run any command → verify `user_id` appears in telemetry spool files
- [ ] `ana logout` → verify credential (including user_id) is cleared
- [ ] Test with existing keyring without user_id → should still work

### DIFF
--- a/src/auth/actions.rs
+++ b/src/auth/actions.rs
@@ -7,9 +7,7 @@ use tokio::time::sleep;
 use super::api_keys::create_api_key;
 use super::errors::AuthError;
 use super::keyring::{delete_api_key, get_api_key, save_credential};
-use super::responses::{
-    AccountResponse, DeviceAuthResponse, OpenIdConfig, TokenErrorResponse, TokenResponse,
-};
+use super::responses::{DeviceAuthResponse, OpenIdConfig, TokenErrorResponse, TokenResponse};
 use crate::context::CommandContext;
 use crate::input::KeyListener;
 use crate::ui::status;
@@ -107,13 +105,10 @@ fn print_token_expiration(expires_at: &str) {
 async fn save_and_display_login(ctx: &CommandContext, api_key: &str) -> Result<(), AuthError> {
     use super::api_keys::get_expiration;
 
-    // Validate the API key against the server before saving
-    let email = validate_api_key(ctx, api_key)
+    // Validate the API key and get user info (email + user_id) in a single request
+    let (email, user_id) = validate_api_key(ctx, api_key)
         .await
         .map_err(|e| AuthError::InvalidApiKey(e.to_string()))?;
-
-    // Fetch user_id for telemetry (best-effort, don't fail login if this fails)
-    let user_id = fetch_user_id(ctx, api_key).await;
 
     // Save to keyring (including user_id if available)
     save_credential(&ctx.config, api_key, user_id.as_deref())?;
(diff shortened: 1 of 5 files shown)


src/auth/actions.rs
```diff
@@ -161,6 +164,31 @@ fn prompt_api_key_hidden() -> Result<String, AuthError> {
     Ok(api_key.trim().to_string())
 }
 
+/// Fetch user_id from the whoami endpoint.
+/// Returns None if the request fails or user_id is not found.
+async fn fetch_user_id(ctx: &CommandContext, api_key: &str) -> Option<String> {
+    let client = ctx.unauthenticated_client(REQUEST_TIMEOUT);
+
+    let response = client
```
mattkram, line 172: We already make a validation request below: https://github.com/anaconda/anaconda-cli/pull/234/changes#diff-d71f63165d18d8c327c7bf175e57d1cec2e13fd428fadcc8c7241ac47e858d35R197

The real intent of that is just to ensure we can make a successful request with the API key. So I think it would be ideal if we could unify the two to prevent multiple requests for account information.

zfralish: Okay, addressed in most revent commit

src/auth/actions.rs
```diff
@@ -164,41 +159,19 @@ fn prompt_api_key_hidden() -> Result<String, AuthError> {
...
 
-/// Fetch user_id from the whoami endpoint.
-/// Returns None if the request fails or user_id is not found.
-async fn fetch_user_id(ctx: &CommandContext, api_key: &str) -> Option<String> {
+/// Validate an API key by making a request to the whoami endpoint.
+/// Returns (email, Option<user_id>) if valid, or an error if invalid.
+async fn validate_api_key(
+    ctx: &CommandContext,
+    api_key: &str,
+) -> miette::Result<(String, Option<String>)> {
```
mattkram, line 169: Is there ever a case where user_id is `None`? Why is the return type `Option<String>`?

zfralish: Only in the case of a service api-key, I think Claude just made it optional in reference to legacy code. Want me to make it a string?

mattkram: Oh, interesting. I think that's fine then, since it could theoretically be None.

src/auth/actions.rs
```diff
@@ -180,19 +181,31 @@ async fn validate_api_key(ctx: &CommandContext, api_key: &str) -> miette::Result
         ));
     }
 
-    let account: AccountResponse = response
+    let data: serde_json::Value = response
```
mattkram: Do we loose anything here by not having a response model for automatic handling of the post-validation response? My assumption was that serde_json should be able to handle some of the unpacking of JSON and validation for us, with less raw dictionary handling below ...

src/auth/actions.rs
```diff
@@ -180,19 +181,31 @@ async fn validate_api_key(ctx: &CommandContext, api_key: &str) -> miette::Result
...
+        .or_else(|| {
+            profile
+                .and_then(|p| p.get("username"))
+                .and_then(|v| v.as_str())
+                .map(|s| s.to_string())
+        })
         .ok_or_else(|| miette::miette!("API key is not associated with a user"))?;
 
-    Ok(email)
+    let user_id = profile
```
mattkram:
```suggestion
    // Some user types (service accounts) may not have a `user_id`, so it is `Option<String>`
    let user_id = profile
```

src/auth/keyring.rs
```diff
@@ -28,10 +28,22 @@ struct Credential {
...
+    user_id: Option<String>,
 }
 
-/// Save an API key to the keyring file.
-pub fn save_api_key(config: &Config, api_key: &str) -> Result<(), AuthError> {
+/// Save an API key to the keyring file (without user_id).
+#[cfg(test)]
+fn save_api_key(config: &Config, api_key: &str) -> Result<(), AuthError> {
+    save_credential(config, api_key, None)
+}
```
mattkram: If this is just for tests, maybe we should update tests? Or move into tests module?

src/auth/keyring.rs
```diff
@@ -50,6 +62,7 @@ pub fn save_api_key(config: &Config, api_key: &str) -> Result<(), AuthError> {
         api_key: api_key.to_string(),
         repo_tokens: vec![],
         version: CREDENTIAL_VERSION,
+        user_id: user_id.map(|s| s.to_string()),
```
mattkram, line 59: Does `serde_json::to_string` store this as literal `null` vs. not including the key? Not sure if it really matters here, but prompting the question in case we want different behavior.

mattkram (approved):
A few nit comments, but overall this looks good. Please feel free to address or decline, I don't think anything is blocking.

PR #236, opened 2026-05-15
ci: Sync release assets with anaconda.sh with each release

marcoesters:
## Summary

Push release assets to `anaconda.sh` with every release. Updates are currently triggered manually. This PR adds an automatic trigger to each release that will perform the update whenever a new release is on GitHub.

Jira: [CLI-630](https://anaconda.atlassian.net/browse/CLI-630)

[CLI-630]: https://anaconda.atlassian.net/browse/CLI-630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -100,3 +100,42 @@ jobs:
 
           echo "Publishing release $VERSION"
           gh release edit "$VERSION" --draft=false --latest
+
+  deploy-to-anaconda-sh:
+    name: Deploy to anaconda.sh
+    if: ${{ github.event_name == 'release' }}
+    runs-on: ubuntu-latest
+    needs: [create-github-release]
+    environment: prod
+    steps:
+      - name: Obtain token
+        id: token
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -100,3 +100,42 @@ jobs:
 
           echo "Publishing release $VERSION"
           gh release edit "$VERSION" --draft=false --latest
+
+  deploy-to-anaconda-sh:
```
mattkram, line 104: Do we need to or can we do anything pre-merge to test this? Or do we need to test during a release?

mattkram (commented):
Overall this looks great. Just one question about testing.

.github/workflows/release.yaml
marcoesters: We cannot test it here, but I tested the code locally with the current token. I will send you the workflow run in private.

mattkram (approved):

PR #239, opened 2026-05-18
fix!: Remove `anaconda` symlink to avoid shadowing in conda's base environment

zfralish:
## Summary
Installing `anaconda-cli` via ana created `~/.ana/bin/anaconda`, which shadowed users' existing `anaconda` command from `anaconda-auth`. This caused issues with `anaconda token` commands.

This PR removes the anaconda symlink entirely. `anaconda-cli` functionality is now only accessible through ana subcommands (e.g., `ana mcp`, `ana org`) rather than a standalone `anaconda` or `anaconda-cli` command.

Changes:
- Set `binaries: &[]` for anaconda-cli in specs.rs (no symlink created)
- Simplified anaconda_cli.rs to use run_tool_binary directly
- Updated mcp/run.rs to check tool_prefix instead of bin_path
- Updated tests to verify no symlink is created

## Test plan
- [ ] Run `ana bootstrap` and verify NO symlink is created in `~/.ana/bin/`
- [ ] Verify `ana mcp` still works (installs and runs anaconda-cli)
- [ ] Verify existing `anaconda` command (from anaconda-auth) is not shadowed

### DIFF
--- a/src/tools/specs.rs
+++ b/src/tools/specs.rs
@@ -16,11 +16,9 @@ const TOOLS: &[Tool] = &[
     Tool {
         name: "anaconda-cli",
         lockfile: include_str!("../../tool-specs/anaconda-cli/pixi.lock"),
-        binaries: if cfg![unix] {
-            &[&["bin", "anaconda"]]
-        } else {
-            &[&["Scripts", "anaconda"]]
-        },
+        // No symlink - anaconda-cli is only accessed via ana subcommands (e.g., ana mcp)
+        // to avoid shadowing users' existing anaconda command from anaconda-auth
+        binaries: &[],
         experimental: None,
     },
     #[cfg(unix)]
--- a/src/anaconda_cli.rs
+++ b/src/anaconda_cli.rs
@@ -1,13 +1,9 @@
-use std::process::Command;
-
 use crate::context::CommandContext;
 use crate::paths;
 use crate::tools;
 
 pub async fn run_bootstrap(ctx: &mut CommandContext) -> Result<(), String> {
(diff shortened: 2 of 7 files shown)


src/tools/specs.rs
```diff
@@ -17,23 +25,35 @@ const TOOLS: &[Tool] = &[
         name: "anaconda-cli",
         lockfile: include_str!("../../tool-specs/anaconda-cli/pixi.lock"),
         binaries: if cfg![unix] {
-            &[&["bin", "anaconda"]]
```
mattkram, line 20: In this case, I think we should just remove this symlink for the anaconda CLI. We can add it again more carefully if needed at a later time.

We should still expose e.g. `ana mcp ...`, but in this case the wrapped subprocess needs to execute `~/.ana/tools/anaconda-cli/bin/anaconda` instead of `~/.ana/bin/anaconda`.

mattkram (commented):
I think for this current issue, we should:

* Just remove the symlink part from the tools.rs for `anaconda-cli` tool (see my other comment)
* Ensure that any wrapped subcommand that calls `anaconda` uses the direct path inside the tool environment, instead of the symlink or shim

For right now, we basically don't want `ana` users to have `anaconda` on path by default. As we continue our roadmap, this could change. But we need to now think more carefully about what the CLI installed-via-conda vs. installed-standalone is, and we need some time to work through that.

mattkram (approved):
@zfralish I pushed some changes to the approach. Mind a second set of eyes and 👍 before I merge?

zfralish (commented):
I think this is better. Thanks!

PR #242, opened 2026-05-19
fix: Use two-step parsing for idiomatic --help handling

mattkram:
## Summary

Alternative approach to #227 that fixes the `--help` panic more idiomatically:

1. **Two-step parsing** - Use `try_get_matches()` + `from_arg_matches()` instead of `try_parse()`, giving us access to raw `ArgMatches` before converting to our typed structs
2. **Derive paths from ArgMatches** - Walk the parsed command tree to build help paths dynamically, eliminating ~110 lines of hardcoded path strings

### Comparison with original fix

| Metric | Original Fix | This Refactor |
|--------|-------------|---------------|
| Files changed | 2 | 4 |
| Net lines | +109 | +113 |
| Churn (adds+removes) | ~84 | ~428 |

The higher churn is mostly restructuring, not new logic. We delete the hardcoded path mappings that would need maintenance every time a subcommand is added.

### Commits

- `fb3e2465` - Zeke's original fix (preserved)
- `0c1a1901` - Refactor to two-step parsing
- `f21247e1` - Replace hardcoded paths with ArgMatches traversal

## Test plan

- [x] `cargo test`
- [x] `cargo clippy`
- [ ] `ana feature enable main-x --help` shows help (not panic)
- [ ] `ana ob app deploy --help` shows help
- [ ] `ana --help` still works

Jira: [CLI-613](https://anaconda.atlassian.net/browse/CLI-613)

[CLI-613]: https://anaconda.atlassian.net/browse/CLI-613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 use std::time::Instant;
 
-use clap::{CommandFactory, Parser, Subcommand};
+use clap::{CommandFactory, FromArgMatches, Parser, Subcommand};
 use miette::{IntoDiagnostic, miette};
 
 use crate::VERSION;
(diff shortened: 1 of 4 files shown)


zfralish (approved):
This feels much better.

Still can't shake the feeling we are missing something with clap though.

I'm really looking for something like the way [`cobra` handles parsing in go,](https://github.com/anaconda/gated-registration-cli/blob/main/cmd/ip.go) I just feel like this amount of manual parsing is abnormal when using a cli lib like `clap`

PR #243, opened 2026-05-19
refac: Replace hand-rolled config with Figment

zfralish:
## Summary

Replaces the hand-rolled environment variable parsing in `Config::load()` with [Figment](https://docs.rs/figment), a layered configuration library. This preserves all existing behavior while laying the groundwork for file-based configuration.

**What changed:**
- Added `figment` dependency with the `env` feature
- Added `Deserialize` to `Config` struct with custom deserializers
- Env vars continue to work identically (`ANA_DOMAIN`, `ANA_SSL_VERIFY=0`, etc.)

**Custom deserializers preserve existing behavior:**
- `deserialize_bool` - Keeps the permissive parsing: empty/0/false → false, anything else → true
- `deserialize_domain` - Normalizes URLs by stripping scheme and path
- `deserialize_self_update_url` - Handles "github" magic value → None

### Why Figment?

The main value isn't code reduction (our custom parsing rules require custom deserializers either way). The value is **layered configuration** - Figment merges multiple config sources in priority order, with later sources overriding earlier ones.

Currently we have:
```
Defaults → Env vars
```

With Figment, adding file-based config becomes trivial:
```rust
Figment::new()
    .merge(Serialized::defaults(Self::defaults()))    // 1. Hardcoded defaults
    .merge(Toml::file("~/.anaconda/config.toml"))     // 2. User config file
    .merge(Toml::file(".ana.toml"))                   // 3. Project config file  
    .merge(Env::prefixed("ANA_"))                     // 4. Env vars (highest priority)
    .extract()
```

This enables:
- **User-level defaults** in `~/.anaconda/config.toml` (e.g., always use a staging domain)
- **Project-level config** in `.ana.toml` checked into repos (e.g., team-specific settings)
- **Env var overrides** still work for CI/scripts, taking highest priority
- **Partial configs** - each layer only needs to specify what it wants to override; Figment handles the merge

Figment also provides better error messages with metadata about which source a value came from, which helps debugging config issues.

## Test plan

- [x] All 201 existing tests pass
- [x] Tests pass with `--features diagnostics`
- [x] `cargo clippy` passes (no new warnings)
- [x] Verified `ana config` shows correct defaults
- [x] Verified env var overrides work: `ANA_DOMAIN=custom.example.com ANA_SSL_VERIFY=0 ana config`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -206,6 +206,7 @@ dependencies = [
  "comfy-table",
  "console",
  "dirs",
+ "figment",
  "futures-util",
  "http 1.4.0",
  "indicatif",
@@ -455,6 +456,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a89cbf775b137e9b968e67227ef7f775587cde3fd31b0d8599dbd0f598a48340"
+dependencies = [
+ "bytemuck",
+]
+
+[[package]]
 name = "atomic-waker"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


zfralish (commented):
**Re: visit_* functions (lines 245-265)**

These visit functions handle different source data types. When serde encounters data, it calls the appropriate visitor based on the actual type:

- `visit_str` - for strings (env vars come through here: `"false"`, `"0"`)
- `visit_bool` - for native booleans (TOML `ssl_verify = true`)
- `visit_i64`/`visit_u64` - for numbers (TOML `ssl_verify = 0`)

Right now with env vars only, everything is a string. But when we add TOML file support, users could write any of:
```toml
ssl_verify = true      # calls visit_bool
ssl_verify = "false"   # calls visit_str  
ssl_verify = 0         # calls visit_i64
```

The multiple visitors let all three forms work correctly.

mattkram (approved):
This is nice. Thanks!

PR #246, opened 2026-05-19
ci: Add secret property to anaconda.sh upload credentials

marcoesters:
## Summary

Add the properties that need to be obtained from the Vault paths for the `anaconda.sh` upload workflow.


mattkram (approved):
That makes sense.

PR #248, opened 2026-05-20
fix: Parse anaconda-auth keyring with embedded repo tokens

mattkram:
## Summary
Fixes #240

The `anaconda-auth` Python library stores credentials in a different format than ana-cli expected:
- `repo_tokens` is a list of objects with `token` and `org_name` fields (not plain strings)
- Uses `username` field instead of `user_id`

This PR updates the keyring credential parsing to be fully compatible with anaconda-auth's `TokenInfo` schema, allowing credentials created by `anaconda token install` to be read by ana-cli.

Changes:
- Add `RepoToken` struct matching anaconda-auth's schema
- Change `repo_tokens` from `Vec<String>` to `Vec<RepoToken>`
- Add `username` field for full schema compatibility
- Fall back to `username` when `user_id` is not set
- Store `username` from whoami response when saving credentials

## Test plan
- [x] All existing tests pass
- [x] New tests for parsing anaconda-auth credential format
- [x] New tests for username fallback behavior
- [ ] Manual test: run `anaconda token install`, then `ana feature enable main-x`

Jira: [CLI-647](https://anaconda.atlassian.net/browse/CLI-647)

[CLI-647]: https://anaconda.atlassian.net/browse/CLI-647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/auth/actions.rs
+++ b/src/auth/actions.rs
@@ -107,17 +107,22 @@ fn print_token_expiration(expires_at: &str) {
 async fn save_and_display_login(ctx: &CommandContext, api_key: &str) -> Result<(), AuthError> {
     use super::api_keys::get_expiration;
 
-    // Validate the API key and get user info (email + user_id) in a single request
-    let (email, user_id) = validate_api_key(ctx, api_key)
+    // Validate the API key and get user info in a single request
+    let user_info = validate_api_key(ctx, api_key)
         .await
         .map_err(|e| AuthError::InvalidApiKey(e.to_string()))?;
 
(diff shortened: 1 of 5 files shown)


zfralish (approved):
I recall username being possibly null, but I am not sure, just something you might want to check in tests if it breaks on a null json value.

PR #250, opened 2026-05-20
feat: Add auth plugin configuration fields to Config

zfralish:
## Summary

[References schema defined here.](https://docs.google.com/document/d/1DqwrFQuCq5lr2pEoqLc3zFb72CuQulC21ue0gyXGQjY/edit?tab=t.0#heading=h.cvg506ocsnkr)

Adds auth plugin configuration support with a multi-source configuration hierarchy:

### Configuration Priority (highest to lowest)
1. **Environment variables** - `ANA_*` and `ANACONDA_AUTH_*` prefixed
2. **Secrets directory** - `/run/secrets` (or `ANACONDA_SECRETS_DIR`) for container secret injection
3. **Dotenv file** - `.env` in current working directory
4. **Config file** - `~/.anaconda/config.toml` `[plugin.auth]` table (or `ANACONDA_CONFIG_TOML`)
5. **Defaults** - Built-in default values

### New Config Fields
- `auth_domain_override` - Override auth domain for OIDC
- `preferred_token_storage` - Token storage location ("system" or "anaconda-keyring")
- `api_key` - Static API key (bypasses OIDC)
- `keyring` - Keyring backend configuration
- `proxy_servers` - Proxy configuration
- `client_cert` / `client_cert_key` - mTLS client certificate paths
- `use_device_flow` - Use OIDC device flow for headless auth

### Changes
- Add figment providers for secrets directory, dotenv, and TOML config file
- Add `dotenvy` dependency for .env parsing
- Add `auth_domain()` helper method
- Update `well_known_url()` to use auth domain override

## Test plan
- [x] All 250 tests pass
- [x] Added 8 new tests for config providers and priority behavior
- [x] `cargo clippy` passes with no errors

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,15 +1,21 @@
 //! Global configuration for ana.
 //!
-//! Configuration is loaded from environment variables. In the future, this will
-//! also support reading from `~/.ana/config.toml`.
+//! Configuration is loaded from multiple sources with the following priority
+//! (highest to lowest):
+//!
+//! 1. **Environment variables** - `ANA_*` and `ANACONDA_AUTH_*` prefixed variables
+//! 2. **Dotenv file** - `.env` file in current working directory
+//! 3. **Config file** - `~/.anaconda/config.toml` `[plugin.auth]` table (or `ANACONDA_CONFIG_TOML`)
+//! 4. **Defaults** - Built-in default values
 //!
 //! # Environment Variables
 //!
 //! | Variable                         | Default                    | Description                     |
 //! |--------------------------------- |----------------------------|---------------------------------|
 //! | `ANA_DOMAIN`                     | `anaconda.com`             | Authentication domain           |
+//! | `ANA_AUTH_DOMAIN_OVERRIDE`       | (none)                     | Override auth domain for OIDC   |
 //! | `ANA_AUTH_CLIENT_ID`             | (Anaconda's ID)            | OAuth client ID                 |
-//! | `ANA_SSL_VERIFY`                 | `true`                     | SSL certificate verification    |
+//! | `ANA_SSL_VERIFY`                 | `true`                     | SSL verification: `true`, `false`, or path to CA cert |
 //! | `ANA_OPEN_BROWSER`               | `true`                     | Auto-open browser during login  |
 //! | `ANA_METRICS_ENDPOINT`           | (Anaconda metrics URL)     | OpenTelemetry metrics endpoint (authenticated) |
 //! | `ANA_METRICS_PUBLIC_ENDPOINT`    | (Anaconda public URL)      | OpenTelemetry metrics endpoint (unauthenticated) |
@@ -21,6 +27,13 @@
 //! | `ANA_PRERELEASES`                | `false`                    | Include prereleases in updates  |
 //! | `ANA_PIP_INDEX_URL`              | `https://repo.anaconda.cloud/repo/anaconda-wheels/simple` | Package index URL for Anaconda wheels |
 //! | `ANA_SELF_UPDATE_URL`            | (Anaconda static URL)      | Update URL; set to `github` for GitHub Releases |
+//! | `ANACONDA_AUTH_AUTH_DOMAIN_OVERRIDE` | (none)                 | Override auth domain for OIDC   |
+//! | `ANACONDA_AUTH_PREFERRED_TOKEN_STORAGE` | `anaconda-keyring`  | Token storage: "system" or "anaconda-keyring" |
+//! | `ANACONDA_AUTH_API_KEY`          | (none)                     | Static API key (bypasses OIDC)  |
+//! | `ANACONDA_AUTH_KEYRING`          | (none)                     | Injected keyring contents (JSON)|
+//! | `ANACONDA_AUTH_PROXY_SERVERS`    | (none)                     | Proxy config (JSON map)         |
+//! | `ANACONDA_AUTH_CLIENT_CERT`      | (none)                     | Client cert path for mTLS       |
+//! | `ANACONDA_AUTH_CLIENT_CERT_KEY`  | (none)                     | Client cert key path for mTLS   |
 //!
 //! When the `diagnostics` feature is enabled:
 //!
@@ -32,13 +45,17 @@
 //! Boolean values are parsed as `false` for empty, "0", or "false" (case-insensitive),
 //! and `true` for any other value.
 
-use std::path::PathBuf;
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
 
-use figment::Figment;
 use figment::providers::{Env, Serialized};
+use figment::value::{Dict, Map, Value};
+use figment::{Error, Figment, Metadata, Profile, Provider};
 use serde::{Deserialize, Deserializer, Serialize};
 
 use crate::table;
+
+const DEFAULT_CONFIG_PATH: &str = ".anaconda/config.toml";
 
 /// Check if telemetry is enabled.
 pub fn telemetry_enabled() -> bool {
@@ -49,7 +66,43 @@ pub fn telemetry_enabled() -> bool {
 
 const DEFAULT_DOMAIN: &str = "anaconda.com";
 const DEFAULT_CLIENT_ID: &str = "b4ad7f1d-c784-46b5-a9fe-106e50441f5a";
-const DEFAULT_SSL_VERIFY: bool = true;
+
+/// SSL verification setting.
+///
+/// Can be a boolean to enable/disable verification, or a path to a custom CA certificate.
+#[derive(Debug, Clone, PartialEq, Serialize)]
+#[serde(untagged)]
+pub enum SslVerify {
+    /// Enable or disable SSL verification
+    Enabled(bool),
+    /// Path to a custom CA certificate file (PEM format)
+    CertPath(PathBuf),
+}
+
+impl Default for SslVerify {
+    fn default() -> Self {
+        SslVerify::Enabled(true)
+    }
+}
+
+impl SslVerify {
+    /// Returns true if SSL verification is enabled (not explicitly disabled).
+    ///
+    /// A cert path is considered "verified" since it adds a custom CA.
+    #[allow(dead_code)] // Will be used when http client is updated
+    pub fn is_verified(&self) -> bool {
+        !matches!(self, SslVerify::Enabled(false))
+    }
+
+    /// Returns the certificate path if one is configured.
+    #[allow(dead_code)] // Will be used when http client is updated
+    pub fn cert_path(&self) -> Option<&Path> {
+        match self {
+            SslVerify::CertPath(p) => Some(p),
+            SslVerify::Enabled(_) => None,
+        }
+    }
+}
 const DEFAULT_OPEN_BROWSER: bool = true;
 const DEFAULT_METRICS_ENDPOINT: &str = "https://metrics.aa.anaconda.com/v1/metrics";
 const DEFAULT_METRICS_PUBLIC_ENDPOINT: &str = "https://public.telemetry.anaconda.com/v1/metrics";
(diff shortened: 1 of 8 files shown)


mattkram (commented):
I'm not sure that we need to propagate all of the values and defaults. We just need to make sure we can read the same config file, and add appropriate handling (and use) for the things that make most sense to us. I think primarily the things we will want is everything around sites and multiple domain support, things like SSL cert paths and proxy servers, etc.

I would recommend we reduce the number of config values we handle. Some of them are only relevant or are things we would want to deprecate anyway.

src/config.rs
```diff
@@ -1,13 +1,20 @@
...
+//! 3. **Dotenv file** - `.env` file in current working directory
+//! 4. **Config file** - `~/.anaconda/config.toml` `[plugin.auth]` table (or `ANACONDA_CONFIG_TOML`)
+//! 5. **Defaults** - Built-in default values
 //!
 //! # Environment Variables
 //!
 //! | Variable                         | Default                    | Description                     |
 //! |--------------------------------- |----------------------------|---------------------------------|
 //! | `ANA_DOMAIN`                     | `anaconda.com`             | Authentication domain           |
+//! | `ANA_AUTH_DOMAIN_OVERRIDE`       | (none)                     | Override auth domain for OIDC   |
```
AlbertDeFusco, line 16: in anaconda-auth this auth-domain override also overrides the keyring look up. Is that true here?

src/config.rs
```diff
@@ -21,6 +28,14 @@
...
 //! | `ANA_PIP_INDEX_URL`              | `https://repo.anaconda.cloud/repo/anaconda-wheels/simple` | Package index URL for Anaconda wheels |
 //! | `ANA_SELF_UPDATE_URL`            | (Anaconda static URL)      | Update URL; set to `github` for GitHub Releases |
+//! | `ANACONDA_AUTH_AUTH_DOMAIN_OVERRIDE` | (none)                 | Override auth domain for OIDC   |
+//! | `ANACONDA_AUTH_PREFERRED_TOKEN_STORAGE` | `anaconda-keyring`  | Token storage: "system" or "anaconda-keyring" |
+//! | `ANACONDA_AUTH_API_KEY`          | (none)                     | Static API key (bypasses OIDC)  |
+//! | `ANACONDA_AUTH_KEYRING`          | (none)                     | Keyring backend config (JSON)   |
+//! | `ANACONDA_AUTH_PROXY_SERVERS`    | (none)                     | Proxy config (JSON map)         |
+//! | `ANACONDA_AUTH_CLIENT_CERT`      | (none)                     | Client cert path for mTLS       |
+//! | `ANACONDA_AUTH_CLIENT_CERT_KEY`  | (none)                     | Client cert key path for mTLS   |
+//! | `ANACONDA_AUTH_USE_DEVICE_FLOW`  | `false`                    | Use OIDC device flow            |
```
AlbertDeFusco: this is now default to true in anaconda-auth

src/config.rs
```diff
@@ -60,18 +80,183 @@ const DEFAULT_USE_HTTPS: bool = true;
...
+            }
+        }
+
+        Ok(Map::from([(Profile::Default, dict)]))
+    }
+}
+
+/// Provider that reads configuration from a .env file.
+///
+/// Supports both `ANA_*` and `ANACONDA_AUTH_*` prefixed variables.
```
AlbertDeFusco, line 124: if both `ANA_FOO` AND `ANACONDA_AUTH_FOO` are set, which takes prececedence?

src/config.rs
```diff
@@ -301,6 +544,31 @@ where
     }
 }
 
+/// Custom deserializer for optional strings that treats empty/whitespace as None.
```
AlbertDeFusco, line 581: in what cases and sources would this get picked up? reading from toml?

src/config.rs
zfralish: It is not used yet, will make a note to make sure it works this way when multi-site is added.

src/config.rs
AlbertDeFusco: i mean that use-device-flow is default to true

src/config.rs
zfralish:
```
  pub fn load() -> Self {
      Figment::new()
          // 5. Defaults (lowest priority)
          .merge(Serialized::defaults(Self::defaults()))
          // 4. Config file
          .merge(TomlConfigProvider::new())
          // 3. Dotenv file
          .merge(DotenvProvider)
          // 2. Secrets directory
          .merge(SecretsDir::new())
          // 1. Environment variables (highest priority)
          .merge(Env::prefixed("ANA_").map(|key| {
              let k = key.as_str().to_lowercase();
              match k.as_str() {
                  "auth_client_id" => "client_id".into(),
                  "prereleases" => "include_prereleases".into(),
                  _ => k.into(),
              }
          }))
          .merge(Env::prefixed("ANACONDA_AUTH_").map(|key| key.as_str().to_lowercase().into()))
          .extract()
          .expect("config loading should not fail with defaults")
  }
```

ANACONDA_AUTH, would take precedence since it is later in the above hierarchy.

AlbertDeFusco: It's possible that while `ana` and `anaconda` co-exist that I might expect `ANA_FOO` to take precedence over `ANACONDA_AUTH_FOO` until we complete migration there is only one `anaconda`. Thoughts?

src/config.rs
zfralish: This would work for any file supported, toml, yaml, env

```
  For example:
  ANACONDA_AUTH_API_KEY=""      # becomes None, not Some("")
  ANACONDA_AUTH_API_KEY="  "    # becomes None, not Some("  ")
```

src/config.rs
zfralish: Right now, I was expecting to not have overlapping vars.

In terms of precedence I am indifferent, and happy to do whatever you recommend.

src/config.rs
zfralish: Handled in [abb55e8](https://github.com/anaconda/anaconda-cli/pull/250/commits/abb55e8e6824e84f88d425f41dab5c0b5c741b9d)

src/config.rs
marcoesters: Since `ANA_*` is newer, I would expect those variables to be deliberately set and thus take precedence over the older `ANACONDA_AUTH_*` variables.

zfralish: Okay, I have made this change [336144d](https://github.com/anaconda/anaconda-cli/pull/250/commits/336144d494ae0e184e55370c95faca8dbf12cf21)

src/config.rs
```diff
@@ -60,24 +117,276 @@ const DEFAULT_USE_HTTPS: bool = true;
...
+        Metadata::named("secrets directory").source(self.path.display().to_string())
+    }
+
+    fn data(&self) -> Result<Map<Profile, Dict>, Error> {
+        if !self.path.is_dir() {
+            return Ok(Map::new());
+        }
+
+        let mut dict = Dict::new();
+        if let Ok(entries) = std::fs::read_dir(&self.path) {
```
marcoesters: This will read all secrets in that directory. I think we should only read the secrets we actually need.

src/config.rs
```diff
@@ -139,10 +473,29 @@ impl Default for Config {
...
+    /// 2. Condarc (`~/.condarc` - select fields only)
+    /// 3. Config file (`~/.anaconda/config.toml` `[plugin.auth]`)
+    /// 4. Dotenv file (`.env` in cwd)
+    /// 5. Secrets directory (`/run/secrets`)
+    /// 6. Environment variables (`ANA_*` and `ANACONDA_AUTH_*`)
     pub fn load() -> Self {
         Figment::new()
+            // 6. Defaults (lowest priority)
             .merge(Serialized::defaults(Self::defaults()))
+            // 5. Condarc (select fields: ssl_verify, proxy_servers, client_cert, client_cert_key)
```
marcoesters: I don't know too much about `anaconda-auth` and all the use cases surrounding that merge chain,  but I am concerned that this will end up being a source of instability, especially if we allow non-ana sources (like the `.condarc` file) to influence `ana`'s behavior. The merging of several disparate sources could also lead to unexpected behavior. If we use `.condarc` file at runtime, it also seems reasonable to add `pixi` and other configuration files to the mix, which would make this even more unpredictable.

I understand that this was done to replicate existing `anaconda-auth` behavior, but I'm wondering if this is the right approach and whether we should be more opinionated on where configuration files are stored for `ana` (either inside `.ana/config.toml` or inside the `anaconda-cli` tool directory).

Similarly, picking up the `.env` file the CWD automatically instead of a user explicitly pointing to a file seems like a recipe for surprising behavior. Maybe I don't see the use cases, but `.env` and `/run/secrets` seem like CI-related convenience wrappers to me. I think requiring the workflow converting those explicitly into `ANA_*` variables seems reasonable for now - this is not an uncommon pattern either.

Using the `~/.anaconda` and `~/.condarc` sources may a be a very nice convenience factor, which could still use to migrate the configuration at install time though.

marcoesters (commented):
I'm not an expert in the authentication workflow, but the scope of this PR seems to have changed quite drastically since the first review comment from supporting these fields to a full hierarchy.

The hierarchy seems to replicate the hierarchy in `anaconda-auth` based on its current architecture (the attached architecture document appears to be three years old). I'm not sure this scales as we add more tools to support.

We have an opportunity to be more opinionated here. However, I am not an expert in this, so I'd be curious to hear @mattkram's thoughts on this.

src/config.rs
zfralish: Speaking strictly on the `.condarc`, we only consider a subset of fields around networking config.

This is largely due to practitioners expecting the condarc to be a one stop shop for all things network config in the conda ecosystem, as I understand it.

marcoesters: Not if they are using `pixi`, which is also part of the conda ecosystem.

src/config.rs
```diff
@@ -60,24 +117,276 @@ const DEFAULT_USE_HTTPS: bool = true;
...
+        Self { path }
+    }
+}
+
+impl Provider for SecretsDir {
+    fn metadata(&self) -> Metadata {
+        Metadata::named("secrets directory").source(self.path.display().to_string())
+    }
+
+    fn data(&self) -> Result<Map<Profile, Dict>, Error> {
```
mattkram: Currently, the only use for the secrets dir is for passing API key into Docker builds. In that scenario, it's only for the conda plugin side of `anaconda-auth` to be able to inject into conda's auth handling mechanism.

I'd recommend we don't read secrets from the Rust implementation side. We can add later if we find it to be necessary.

src/config.rs
```diff
@@ -60,24 +117,276 @@ const DEFAULT_USE_HTTPS: bool = true;
...
+        }
+        toml::Value::Datetime(dt) => Value::from(dt.to_string()),
+    }
+}
+
+/// Provider that reads select configuration from `~/.condarc`.
+///
+/// Only reads fields relevant to ana: `ssl_verify`, `proxy_servers`,
+/// `client_cert`, and `client_cert_key`.
+struct CondarcProvider {
```
mattkram: We did this in anaconda-auth, primarily due to the conda plugin and the difficulty in differentiating between it, and conda itself. And resulting customer/user confusion.

We may find we need it here too. But I'm hesitant to want to keep it right now, as raw YAML parsing etc. and loading conda's configuration are brittle and I'm not sure the extra complexity is justified.

src/config.rs
```diff
@@ -118,6 +427,31 @@ pub struct Config {
...
 
+    /// Where to store authentication tokens: "system" or "anaconda-keyring"
+    pub preferred_token_storage: String,
+
+    /// Static API key for authentication (bypasses OIDC flow when set)
+    #[serde(default, deserialize_with = "deserialize_optional_string")]
+    pub api_key: Option<String>,
+
+    /// Keyring backend configuration (backend identifier -> settings)
+    pub keyring: Option<HashMap<String, HashMap<String, String>>>,
```
mattkram, line 303: Both of these really are intended to be used only as environment variables. Is there a way to prevent or discourage storing them in `config.toml`?

src/config.rs
```diff
@@ -118,6 +427,31 @@ pub struct Config {
...
+    #[serde(default, deserialize_with = "deserialize_optional_string")]
+    pub client_cert: Option<String>,
+
+    /// Path to private key for client certificate
+    #[serde(default, deserialize_with = "deserialize_optional_string")]
+    pub client_cert_key: Option<String>,
+
+    /// Use OIDC Device Authorization Grant flow instead of Authorization Code flow
+    #[serde(deserialize_with = "deserialize_bool")]
+    pub use_device_flow: bool,
```
mattkram: We should just always use device flow. I'm not intending us to do anything but that in Rust, and we just changed the default in `anaconda-auth` too.

src/config.rs
```diff
@@ -118,6 +427,31 @@ pub struct Config {
     #[serde(deserialize_with = "deserialize_self_update_url")]
     pub self_update_url: Option<String>,
 
+    /// Where to store authentication tokens: "system" or "anaconda-keyring"
+    pub preferred_token_storage: String,
```
mattkram, line 296: This one might cause confusion until we implement system keyring support. We may want to warn if users have it set to `system` but use Rust for login.

src/config.rs
zfralish: Should be handled [231cb7a](https://github.com/anaconda/anaconda-cli/pull/250/commits/231cb7a471e705d3cf286a4c4804f8aaa715e9ec)

src/config.rs
zfralish: [dc3cadf](https://github.com/anaconda/anaconda-cli/pull/250/commits/dc3cadf0aa872fcb9198beb52adf64bd8bc3f6cb)

src/config.rs
zfralish: [10fb9e8](https://github.com/anaconda/anaconda-cli/pull/250/commits/10fb9e8dc245ff3fc89d08f707bbe260fb256dba)

src/config.rs
zfralish: [be3366e](https://github.com/anaconda/anaconda-cli/pull/250/commits/be3366ee118172358aef2efaab768d7ef3a44048)

src/config.rs
zfralish: [814ca24](https://github.com/anaconda/anaconda-cli/pull/250/commits/814ca24354594b36beb49ffc4bd27516b5010797)

src/config.rs
```diff
@@ -21,6 +27,13 @@
 //! | `ANA_PRERELEASES`                | `false`                    | Include prereleases in updates  |
 //! | `ANA_PIP_INDEX_URL`              | `https://repo.anaconda.cloud/repo/anaconda-wheels/simple` | Package index URL for Anaconda wheels |
 //! | `ANA_SELF_UPDATE_URL`            | (Anaconda static URL)      | Update URL; set to `github` for GitHub Releases |
+//! | `ANACONDA_AUTH_AUTH_DOMAIN_OVERRIDE` | (none)                 | Override auth domain for OIDC   |
+//! | `ANACONDA_AUTH_PREFERRED_TOKEN_STORAGE` | `anaconda-keyring`  | Token storage: "system" or "anaconda-keyring" |
+//! | `ANACONDA_AUTH_API_KEY`          | (none)                     | Static API key (bypasses OIDC)  |
+//! | `ANACONDA_AUTH_KEYRING`          | (none)                     | Keyring backend config (JSON)   |
```
AlbertDeFusco: I wouldn't say this is a "backend config" but the contents of the anaconda-keyring itself.

```text
❯ jq . ~/.anaconda/keyring
{
  "Anaconda Cloud": {
    "anaconda.com": "<redacted-b64-encoded-payload>",
    "my-onprem.local": "
<redacted-b64-encoded-payload>"
  }
}
```

zfralish: [aee94bc](https://github.com/anaconda/anaconda-cli/pull/250/commits/aee94bc28fda34f69a8397619464f88d35b704a9)

PR #259, opened 2026-05-26
fix: Reject empty URL in ana fetch command

zfralish:
## Summary
- Add validation to reject empty URL strings in `ana fetch` command
- Returns a clear error message "URL cannot be empty" instead of making an invalid request

## Test plan
- [ ] Run `ana fetch ""` and verify it returns error "URL cannot be empty"
- [ ] Run `ana fetch /api/auth/passport` and verify it still works normally

Jira: [CLI-628](https://anaconda.atlassian.net/browse/CLI-628)

[CLI-628]: https://anaconda.atlassian.net/browse/CLI-628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


marcoesters (commented):
This entire file could use unit tests to handle different errors, but at least the new code should be covered.

src/fetch.rs
```diff
@@ -46,3 +50,86 @@ pub async fn api_fetch(
     println!("{}", body);
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
```
marcoesters, line 55: These tests are circular and don't test the actual code since you only test the functions that you define within the `test` modules. If it's too complicated for this PR to create or use a mocker for the client object, I'm happy to defer this to a different PR as long as an issue is created for it.

zfralish: Yeah sorry, been letting claude chew on these bug prs. Will either push a fix or defer to another pr.

src/fetch.rs
```diff
@@ -46,3 +50,220 @@ pub async fn api_fetch(
...
+    use std::path::PathBuf;
+
+    use wiremock::matchers::{body_json, header, method, path, query_param};
+    use wiremock::{Mock, MockServer, ResponseTemplate};
+
+    use super::*;
+    use crate::config::Config;
+    use crate::http::Client;
+
+    fn test_config(keyring_path: PathBuf, domain: &str) -> Config {
```
zfralish, line 65: Could probably centralize this somewhere, but atomic has its uses too.

src/fetch.rs
zfralish: Okay, got actual tests now, ballooned the scope a bit due to having to make the context client object mockable, but will be useful in the future.

marcoesters: I like this, now we have unit tests for this public function - thank you!

marcoesters (approved):

src/fetch.rs
zfralish: Yeah I am going to follow up with a refactor pr against the entire context object, make the entire thing easily mockable, everything will be easier to test that way.

PR #266, opened 2026-06-02
fix: Auto-update managed tools when lockfile changes

mattkram:
## Summary
- Add `ensure_tool()` function that compares embedded lockfile hash with stored hash to detect when updates are needed
- Add `ana tool update` command to update all installed managed tools
- After `ana self update`, automatically spawn the new binary to update installed tools with new lockfiles
- Store lockfile hash in `.lockfile-hash` file within each tool's prefix directory
- Add per-tool `auto_update` setting in tool specs (anaconda-cli and outerbounds default to true, pixi defaults to false)
- Add `ANA_AUTO_UPDATE_TOOLS` env var to globally enable/disable auto-update, overriding per-tool defaults

This fixes the issue where `ana self update` would leave managed tools outdated, causing commands like `ana mcp setup` to fail silently when they required newer tool versions.

The core assumption for this behavior is that the tools managed by `ana` are completely defined by the versions in their corresponding lockfile. This is appropriate given the current assumptions of the way we do tool management. We may relax this in the future, but that will introduce more complexity around updates that is not necessary at this time.

### Auto-update configuration

| `ANA_AUTO_UPDATE_TOOLS` | Behavior |
|-------------------------|----------|
| `true` | All installed tools auto-update on `ana self update` |
| `false` | No tools auto-update |
| unset/empty | Each tool uses its own default (`anaconda-cli`, `outerbounds`: update; `pixi`: skip) |

## Test plan
- [x] `cargo clippy` passes
- [x] `cargo test` passes
- [x] Integration tests added for `ana tool update`
- [ ] Manual test: corrupt hash file and verify `ana tool update` triggers reinstall
- [ ] Windows CI passes

Jira: [CLI-623](https://anaconda.atlassian.net/browse/CLI-623)

[CLI-623]: https://anaconda.atlassian.net/browse/CLI-623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/telemetry/otel.rs
```diff
@@ -113,7 +113,7 @@ mod tests {
     fn test_serializable_value_from_otel() {
         let string_val = Value::String("test".into());
         let int_val = Value::I64(42);
-        let float_val = Value::F64(3.14);
+        let float_val = Value::F64(2.72);
```
mattkram: These are harmless but unrelated linter fixes I decided to leave in.

src/tools/install.rs
```diff
@@ -23,6 +23,58 @@ static MULTI_PROGRESS: std::sync::LazyLock<MultiProgress> = std::sync::LazyLock:
...
+    let prefix = paths::tool_prefix(name);
+    let hash_file = prefix.join(".lockfile-hash");
+
+    let lock_content =
+        specs::content(name).ok_or_else(|| miette::miette!("unknown tool: {}", name))?;
+    let current_hash = hash_lockfile(&lock_content);
+
+    // Check if tool is installed and lockfile hash matches
+    if prefix.exists()
+        && let Ok(stored_hash) = std::fs::read_to_string(&hash_file)
```
marcoesters, line 73: This would imply that after this change, all tools will try to update at least once (since old versions of `anaconda-cli` do not have the hash file).

src/update.rs
```diff
@@ -466,7 +506,10 @@ pub async fn run_update(
 
         let start = std::time::Instant::now();
         match apply_update(ctx, &release).await {
-            Ok(()) => print_update_success(current_version, &release.tag_name, start.elapsed()),
+            Ok(()) => {
+                print_update_success(current_version, &release.tag_name, start.elapsed());
+                update_installed_tools();
```
marcoesters: Shouldn't the order be flipped? This reads like we're printing the success message before the update.

src/update.rs
```diff
@@ -413,6 +414,45 @@ fn print_update_success(current_version: &str, new_version: &str, elapsed: std::
     eprintln!();
 }
 
+/// Spawn the new binary to update installed tools.
+///
+/// After self-replace, the current process still has old lockfiles embedded.
+/// We spawn the new binary to update tools using the new lockfiles.
```
marcoesters, line 407: This would mean that updating `ana` would also update all installed tools. I'm not sure that's desired.

src/tools/install.rs
mattkram: Yes. In those cases it would either be an upgrade, or basically a no-op.

src/update.rs
mattkram: Yes, that was my intention. Do you think we should do that differently?

I think, since right now the tools are designed more-or-less (anaconda-cli and outerbounds in particular) to be wrapped directly by ana, this is appropriate. I could see this being more configurable for other tools in the future.

mattkram: I have added some configurability here. Tools can be declared with their own "auto-update" default, and users can set a global override for this. 

`anaconda-cli` and `outerbounds` tools will be auto-updated by default, pixi won't. We will probably expand this out a bit more based on feedback in the future.

zfralish (approved):

PR #273, opened 2026-06-08
feat: Add ability to download a Miniconda installer

ericdill:
`ana download miniconda` — grabs the latest miniconda installer for your platform, checksums it, drops it in your cwd, prints the command to run it. it does not run the installer.

```
$ ana download miniconda
Downloaded Miniconda3-latest-MacOSX-arm64.sh (127.4 MB) to:
    ./Miniconda3-latest-MacOSX-arm64.sh
SHA256 verified.

To install, run:
    bash ./Miniconda3-latest-MacOSX-arm64.sh
```

context: i got a new laptop, wanted miniconda, didn't want to open a browser and go hunt down the right .sh for my arch. that's the whole thing. that's all this solves.

why not `ana install miniconda` that actually runs it? because the moment you run the installer you're in a totally different conversation.
- `-b` is literally license acceptance, so now ana is accepting the TOS on the user's behalf and that's another place we're maintaining a legal position.
- and you're stuck either masking miniconda's own UX behind our defaults or just re-exposing all its flags.

download-only sidesteps all of it. user runs it themselves, accepts the terms interactively, or uses cli flags, same as they would today.

what's actually in here:

- new `src/installer/` module. did NOT route it through `src/tools/` — that's rattler/lockfile/symlink machinery and this shares none of it, it's just "download a file and verify it"
- platform detection → filename for the 5 targets we actually ship (mac arm64/x86_64, linux x86_64/aarch64, win x86_64). anything else is a clean error, not a guessed URL that 404s
- sha256 verify against `repo.anaconda.com/miniconda/.files.json`. fails closed — key missing or empty checksum in the manifest and we refuse, we don't hand you something we couldn't verify
- temp file + atomic rename so a bad download never leaves a half-written file behind
- file already in your cwd → insta-fail ("remove it if you want to continue"), checked *before* the download so we don't pull 120mb just to bail at the end

stuff i left out on purpose — these are follow-ons, not oversights:

- `-f`/`--force` to overwrite instead of failing. obvious next flag. and miniconda's own installer uses `-f` for "no error if prefix exists" so it lines up nicely
- `--output PATH` to land it somewhere other than cwd. this is the one that makes it useful in CI: `ana download miniconda --output /tmp/mc.sh && bash /tmp/mc.sh -b -p /tmp/miniconda`
- not hardcoding `repo.anaconda.com` forever. we already config-drive our other remotes (`ANA_DOMAIN`, `ANA_PIP_INDEX_URL`) and the installer host should resolve the same way eventually. env override, then logged-in customer's host, then public default. it's not immediately clear to me how the custom installer stuff that our proserv team manages works. we should probably have some thinking about how to better support simpler bootstrapping with custom installers. but that's out of scope for this. also matters for air-gapped customers who can't reach `repo.anaconda.com` at all? does PSM mirror the installers too? if so we could look at supporting that use case too.
- other distributions / pinned versions. anaconda distribution lives at `/archive/.files.json`, no `latest` alias so that one needs an explicit version

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -242,7 +242,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml",
- "sha2",
+ "sha2 0.10.9",
  "temp-env",
  "tempfile",
  "thiserror 2.0.18",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -242,6 +242,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml",
+ "sha2",
  "temp-env",
  "tempfile",
  "thiserror 2.0.18",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ self-replace = "1.5"
 sentry = { version = "0.48", optional = true }
 semver = "1.0"
 serde = { version = "1.0", features = ["derive"] }
+sha2 = "0.10"
 serde_json = "1.0"
 serde_yaml = "0.9"
 toml = "1.0"
--- a/docs/design/tool-download.md
+++ b/docs/design/tool-download.md
@@ -1,0 +1,288 @@
+# Implementation Plan: `ana tool download miniconda`
+
+**Status:** Ready to build — Jun 8, 2026. **Repo:** `anaconda/anaconda-cli`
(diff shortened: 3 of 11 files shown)


mattkram (approved):
Looks like a great simple feature to add. I've refactored slightly to re-use the progress bar construction for consistency.

PR #279, opened 2026-06-11
ci: Add bot code review workflow

Umanan23:
## Summary
Add GitHub Actions workflow to enable @claude bot PR reviews. This workflow delegates PR reviews to the org's reusable workflow in internal-actions, matching the implementation used in ai-core.

The workflow triggers on:
- New/reopened pull requests
- Issue comments, PR review comments, and PR reviews that mention `@claude`

## Test plan
- [ ] Create a test PR and verify the workflow runs
- [ ] Comment with `@claude` on a PR and verify the bot responds

Jira: [CLI-684](https://anaconda.atlassian.net/browse/CLI-684)

[CLI-684]: https://anaconda.atlassian.net/browse/CLI-684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/bot-code-review.yaml
+++ b/.github/workflows/bot-code-review.yaml
@@ -10,7 +10,7 @@ on:
     types: [submitted]
   pull_request:
     types: [opened, reopened]
-
+permissions: {}
 jobs:
   claude-code-review:
     name: Claude Code Review
--- a/.github/workflows/bot-code-review.yaml
+++ b/.github/workflows/bot-code-review.yaml
@@ -15,10 +15,10 @@ jobs:
   claude-code-review:
     name: Claude Code Review
     permissions:
-      contents: read
-      pull-requests: write
-      id-token: write
-      actions: read
+      contents: read # To read the repo and PR
+      pull-requests: write # To leave comments on the PR
+      id-token: write # Needed for oidc assume IAM role
+      actions: read # Needed to identify CI failures
     if: |
       (github.event_name == 'pull_request') ||
       (github.event_name == 'issue_comment' && github.event.issue.pull_request && contains(github.event.comment.body, '@claude')) ||
--- a/.github/workflows/bot-code-review.yaml
+++ b/.github/workflows/bot-code-review.yaml
@@ -24,6 +24,6 @@ jobs:
       (github.event_name == 'issue_comment' && github.event.issue.pull_request && contains(github.event.comment.body, '@claude')) ||
       (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@claude')) ||
       (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@claude'))
-    uses: anaconda/internal-actions/.github/workflows/bot-code-review.yml@1be176d1d4801c4ca716ffaf646726200b60e82f # main
+    uses: anaconda/internal-actions/.github/workflows/bot-code-review.yml@main
     with:
       team: ai-services
--- a/.github/workflows/bot-code-review.yml
+++ b/.github/workflows/bot-code-review.yml
@@ -1,0 +1,29 @@
+# Delegates @claude PR reviews to the org reusable workflow (internal-actions).
+name: '[Bot] Code Review'
+
+on:
+  issue_comment:
+    types: [created]
+  pull_request_review_comment:
+    types: [created]
+  pull_request_review:
+    types: [submitted]
+  pull_request:
+    types: [opened, reopened]
+
+jobs:
+  claude-code-review:
+    name: Claude Code Review
+    permissions:
+      contents: read
+      pull-requests: write
+      id-token: write
+      actions: read
+    if: |
+      (github.event_name == 'pull_request') ||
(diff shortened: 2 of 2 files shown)


.github/workflows/bot-code-review.yml
```diff
@@ -0,0 +1,29 @@
...
+    types: [opened, reopened]
+
+jobs:
+  claude-code-review:
+    name: Claude Code Review
+    permissions:
+      contents: read
+      pull-requests: write
+      id-token: write
+      actions: read
```
marcoesters: Those permissions need explanatory comments, especially the write permissions.

.github/workflows/bot-code-review.yaml
```diff
@@ -0,0 +1,29 @@
...
+on:
+  issue_comment:
+    types: [created]
+  pull_request_review_comment:
+    types: [created]
+  pull_request_review:
+    types: [submitted]
+  pull_request:
+    types: [opened, reopened]
+
```
marcoesters: The permissions of this workflow are too broad. `zizmor` will complain:

```
warning[excessive-permissions]: overly broad permissions
  --> /Users/mesters/Downloads/bot-code-review.yml:2:1
   |
 2 | / name: '[Bot] Code Review'
 3 | |
 4 | | on:
 5 | |   issue_comment:
...  |
28 | |     with:
29 | |       team: ai-services
   | |________________________^ default permissions used due to no permissions: block
   |
   = note: audit confidence → Medium
```

.github/workflows/bot-code-review.yaml
marcoesters, line 1: All other workflows use `.yaml` as the file ending - we should be consistent.

marcoesters (commented):
I'm not sure why the `zizmor` pre-commit check didn't capture those - maybe because those are new workflows, but that check would have failed on this workflow.

.github/workflows/bot-code-review.yaml
BenHildebrand: @marcoesters  Can you elaborate or provide a fix?

marcoesters: When permissions are not set, the workflow will inherit default token permissions of the repository. It is generally advisable to explicitly set permission on the workflow level.

```YAML
permissions: {}
```

at the workflow level should do it here. That would satisfy the linting rule.

BenHildebrand (approved):

marcoesters (commented):
I see that this workflow is still failing: https://github.com/anaconda/anaconda-cli/actions/runs/27373820460

What's the origin of those failures? It looks like it cannot access the private action.

.github/workflows/bot-code-review.yaml
marcoesters: Adding another comment to see if the action works.

.github/workflows/bot-code-review.yaml
```diff
@@ -0,0 +1,29 @@
...
+      contents: read # To read the repo and PR
+      pull-requests: write # To leave comments on the PR
+      id-token: write # Needed for oidc assume IAM role
+      actions: read # Needed to identify CI failures
+    if: |
+      (github.event_name == 'pull_request') ||
+      (github.event_name == 'issue_comment' && github.event.issue.pull_request && contains(github.event.comment.body, '@claude')) ||
+      (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@claude')) ||
+      (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@claude'))
+    uses: anaconda/internal-actions/.github/workflows/bot-code-review.yml@main
```
marcoesters, line 27:
```suggestion
    uses: anaconda/internal-actions/.github/workflows/bot-code-review.yml@fc6dd92cbb6930b626625b10c4aafade6c8c5452  # v1.6.0
```

Pinning to a hash is, in my opinion, a hard requirement. This isn't just bad practice from a supply-chain perspective, it will also prevent that upstream issues propagate into pull requests here.

I pinned it to the latest version here, but I'm open to discussing pinning to `main` (as long as it's hashed).

marcoesters (commented):
Even when pinning to `main`, the workflow is still not working. See these two runs that triggered after I responded to a [comment](https://github.com/anaconda/anaconda-cli/pull/279#discussion_r3399506120):

* https://github.com/anaconda/anaconda-cli/actions/runs/27380933165
* https://github.com/anaconda/anaconda-cli/actions/runs/27380933802

It's possible that it's because we are trying to access a private repository from a public repository here. I also don't know why one comment triggered two workflow runs.

mattkram (commented):
I'm taking a quick look. I'm not sure if we will be able to use the private action in a public repository. @BenHildebrand have you seen any usages elsewhere where this is not the case?

Umanan23:
@BenHildebrand  Should i then Close this PR on behalf of your above comment? But are we still considering to add the bot file after brainstorming?

PR #291, opened 2026-06-17
feat: Add externally-managed installers table to `ana tool list`

ericdill:
## What

`ana tool list` now prints a **second table** for externally-managed installers — things `ana` downloads (and SHA256-verifies) but does *not* manage afterward — keeping them distinct from the rattler-managed tools in the first table.

```
Externally Managed Installers
┌───────────┬─────────┬─────────────────────────────┬─────────────┐
│ Name      │ Version │ Download command            │ Status      │
╞═══════════╪═════════╪═════════════════════════════╪═════════════╡
│ miniconda │ latest  │ ana tool download miniconda │ available   │
├───────────┼─────────┼─────────────────────────────┼─────────────┤
│ anaconda  │ n/a     │ n/a                         │ coming soon │
└───────────┴─────────┴─────────────────────────────┴─────────────┘
```

The `anaconda` row is intentionally a roadmap placeholder (no implementation behind it yet) — it's there to surface demand. See telemetry change below.

## Why a second table

Miniconda is categorically different from `pixi`/`outerbounds`/`anaconda-cli`: those are rattler-managed conda environments installed into `~/.ana/tools` with symlinked binaries. `ana tool download miniconda` instead fetches the `Miniconda3-latest-*` installer into the cwd, verifies it, and leaves installation to the user. Conflating the two in one table would misrepresent what `ana` manages.

## Telemetry: capture demand for unsupported installers

`ana tool download <name>` now records the **name the user typed** as an `installer_name` telemetry attribute. The unsupported-name path was moved out of arg-parsing (which `process::exit`ed before any telemetry context existed) and into the command handler, where it returns an error — so the attempt is recorded on the `cli_command_failure` counter under `command: tool.download`.

This means if a user sees `anaconda` "coming soon" and runs `ana tool download anaconda`, we get a measurable signal — demand attempted, not just demand spoken — to decide whether to build it.

## Test plan

- [x] `cargo test` — 248 passing
- [x] `ana tool list` renders both tables (shown above)
- [x] `ana tool download anaconda` → `only miniconda is currently supported (got 'anaconda')`, exit 1, spools `installer_name: anaconda`
- [x] `ana tool download` (no name) → subcommand help

[CLI-700](https://anaconda.atlassian.net/browse/CLI-700)

[CLI-700]: https://anaconda.atlassian.net/browse/CLI-700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -165,7 +165,9 @@ pub enum Action {
         pixi: bool,
     },
     FeatureList,
-    DownloadMiniconda,
+    ToolDownload {
+        name: String,
+    },
     TelemetrySubmit,
     TelemetryKill,
     TelemetryStatus,
@@ -211,7 +213,7 @@ impl Action {
                 _ => "feature.disable.unknown",
             },
             Action::FeatureList => "feature.list",
-            Action::DownloadMiniconda => "tool.download.miniconda",
+            Action::ToolDownload { .. } => "tool.download",
             Action::TelemetrySubmit => "telemetry-submit",
             Action::TelemetryKill => "telemetry-kill",
             Action::TelemetryStatus => "telemetry-status",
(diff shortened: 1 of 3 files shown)


src/cli.rs
```diff
@@ -481,7 +483,18 @@ impl Action {
                 feature::list::print_feature_list(ctx);
                 Ok(())
             }
-            Action::DownloadMiniconda => installer::run(ctx, None).await,
+            Action::ToolDownload { name } => {
+                // Record what the user actually typed so we can see demand for
+                // installers we don't support yet (e.g. "anaconda").
+                ctx.telemetry.add("installer_name", name.clone());
```
ericdill, line 489: @mattkram this gives us telemetry on what users are actually typing here. its another good demand signal for what they might be wanting / expecting.

mattkram: I like it.

mattkram (approved):
Looks great. Just added a quick integration test.

PR #298, opened 2026-06-22
fix: Clarify shell restart needed after installation

redeux:
## Summary
After install, users are told to run `ana --help` but the command won't work until they restart their shell. This updates the final message to include restart instructions when the shell profile was modified.

Fixes #277


mattkram (approved):
Thanks!

PR #311, opened 2026-06-25
feat: Add conda-package feature for distribution as conda package

mattkram:
## Summary
Adds feature flags and build infrastructure to support distributing `ana` as a conda package:

- **`conda-package` feature**: When enabled, disables self-managed functionality (self-update, tool install/uninstall) since these are managed by conda
- **`self-update` feature**: Makes `ana self update` optional (included by default, disabled for conda builds)
- **`tool-install` feature**: Makes `ana tool install/uninstall` optional with rattler as an optional dependency (included by default, disabled for conda builds)
- **cfg-aliases**: Uses `cfg_aliases` crate for cleaner feature flag handling in the codebase
- **anaconda-mcp check**: Checks for `anaconda-mcp` availability before running MCP subcommand rather than hard-depending on it
- **Recipe updates**: Updated conda recipe to use the new feature flags

When built with `--no-default-features --features conda-package`:
- `ana self update` returns an error explaining it's not available
- `ana tool install/uninstall` returns an error explaining tools are managed by conda
- `ana tool list` still works (shows available tools)
- `ana bootstrap` is a no-op (anaconda-cli provided by conda)
- Binary resolution uses `$CONDA_PREFIX/bin/` instead of internal tool paths

## Test plan
- [ ] `cargo build` - default features work
- [ ] `cargo build --no-default-features --features conda-package` - conda package build works
- [ ] `cargo test` - all tests pass
- [ ] `cargo clippy` - no warnings

### DIFF
--- a/src/mcp/run.rs
+++ b/src/mcp/run.rs
@@ -1,15 +1,15 @@
-#[cfg(feature = "conda-package")]
+#[cfg(not(tool_install))]
 use std::path::PathBuf;
 
 use crate::context::CommandContext;
-#[cfg(not(feature = "conda-package"))]
+#[cfg(tool_install)]
 use crate::paths;
 use crate::tools;
-#[cfg(not(feature = "conda-package"))]
+#[cfg(tool_install)]
 use crate::ui::status;
 
 /// Check if anaconda-mcp is installed by looking for its conda-meta entry.
-#[cfg(feature = "conda-package")]
+#[cfg(not(tool_install))]
 fn is_anaconda_mcp_installed() -> bool {
     let Some(conda_prefix) = std::env::var("CONDA_PREFIX").ok() else {
         return false;
@@ -38,7 +38,7 @@ fn is_anaconda_mcp_installed() -> bool {
 /// When built without `conda-package` feature, auto-installs anaconda-cli if not present.
 /// When built with `conda-package` feature, anaconda-cli is expected to be provided by conda,
 /// and anaconda-mcp must be installed for the mcp subcommand to work.
-#[cfg(feature = "conda-package")]
+#[cfg(not(tool_install))]
 pub async fn run(_ctx: &mut CommandContext, args: &[String]) -> miette::Result<()> {
     if !is_anaconda_mcp_installed() {
         return Err(crate::errors::AnacondaMcpNotInstalledError.into());
@@ -49,7 +49,7 @@ pub async fn run(_ctx: &mut CommandContext, args: &[String]) -> miette::Result<(
     tools::run_tool_binary("anaconda-cli", "anaconda", &mcp_args)
 }
 
-#[cfg(not(feature = "conda-package"))]
+#[cfg(tool_install)]
 pub async fn run(ctx: &mut CommandContext, args: &[String]) -> miette::Result<()> {
     if !paths::tool_prefix("anaconda-cli").exists() {
         status::info("Installing anaconda-cli...");
--- a/src/mcp/run.rs
+++ b/src/mcp/run.rs
@@ -1,10 +1,22 @@
 use crate::context::CommandContext;
+#[cfg(not(feature = "conda-package"))]
 use crate::paths;
 use crate::tools;
+#[cfg(not(feature = "conda-package"))]
 use crate::ui::status;
 
 /// Run the `anaconda mcp` command with the given arguments.
-/// Auto-installs anaconda-cli if not present.
+///
+/// When built without `conda-package` feature, auto-installs anaconda-cli if not present.
+/// When built with `conda-package` feature, anaconda-cli is expected to be provided by conda.
+#[cfg(feature = "conda-package")]
+pub async fn run(_ctx: &mut CommandContext, args: &[String]) -> miette::Result<()> {
+    let mut mcp_args = vec!["mcp".to_string()];
+    mcp_args.extend(args.iter().cloned());
+    tools::run_tool_binary("anaconda-cli", "anaconda", &mcp_args)
+}
+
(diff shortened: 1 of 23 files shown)


mcg1969:
I asked Grok to review this, and filtered out the issues I felt weren't relevant.

**Update notifier still points at `ana self update`**
`check_for_update_notification` / `update_notifier::show_notification` are not gated on `self_update`. Conda-package users will still get “Update with: `ana self update`”, then a hard error if they follow it. Gate the notifier the same way as the command (and ideally `cfg` out `update_notifier` entirely).

**Cfg mismatch on `AnacondaMcpNotInstalledError`**
The type is `#[cfg(feature = "conda-package")]` but it's used under `#[cfg(not(tool_install))]`. `cargo build --no-default-features` (without `conda-package`) should fail to compile. Align both on `not(tool_install)` or `feature = "conda-package"`.

**MCP check vs hard run dep**
The recipe always installs `anaconda-mcp`, but `mcp/run.rs` still special-cases a missing conda-meta entry. Either keep MCP optional and keep the check, or drop the check.

**`CONDA_PREFIX` is required**
`resolve_anaconda_bin` / `resolve_tool_binary` die if `CONDA_PREFIX` is unset. Invoking `ana` by absolute path without an activated env is a common conda path. Prefer deriving the prefix from `current_exe()`, then fall back to `CONDA_PREFIX`.

**`ana tool list` still looks at `~/.ana/tools/`**
In the conda build this will mark everything uninstalled and still print miniconda download commands that don't work. Either teach list about `$CONDA_PREFIX` or don't present list as a working conda-mode command.

**`ana ob` is compiled out** of the conda binary (`all(unix, tool_install)`). Confirm that's intended.

**Nits**
- CONDA_PREFIX lookup is duplicated in `anaconda_cli.rs` and `tools/run.rs`; the former also reimplements process spawn instead of `run_tool_binary`.
- `self-replace` and the `update` module still compile into the conda binary (`allow(dead_code)` rather than `cfg`).
- Tests cover the happy smoke path; missing coverage for unset `CONDA_PREFIX`, `ana org` through the proxied binary, MCP missing, and a CI `cargo check --no-default-features --features conda-package` (pytest may build that binary, unit tests do not).

zfralish:
Thanks for the review @mcg1969 — addressed in 1f863d2. Details:

**Fixed (real bugs)**
- **Update notifier** — now gated behind `self_update`; conda users will no longer be told to run `ana self update`. The whole `update_notifier` module is compiled out of conda builds.
- **Cfg mismatch** — `AnacondaMcpNotInstalledError` aligned to `not(tool_install)`; `cargo check --no-default-features` compiles again (confirmed it was broken).

**Fixed (valid improvements)**
- **`CONDA_PREFIX` requirement** — new `paths::conda_prefix()` derives the prefix from `current_exe()` (validated by checking for `conda-meta/`), falling back to `$CONDA_PREFIX`. This also deduplicated the three copies of the lookup. Non-conda builds are untouched — the code is entirely `cfg`'d out.
- **`tool list` in conda mode** — installation status now comes from the env's `conda-meta` entries, so `anaconda-cli` shows ✓ when installed via conda. (The miniconda download table stays — `ana tool download` is not gated and works in conda builds.)
- **CI coverage** — new step runs `cargo check --no-default-features` and `cargo test --features conda-package` in an isolated `target/conda-package/` dir so it can't clobber the release binary the integration tests rely on.

**No change (intentional)**
- **MCP guard stays** — `anaconda-mcp` is a hard run dep again, but the conda-meta check remains as a defensive guard for envs where the package was removed.
- **`ana ob` stays compiled out** — outerbounds auto-installs via rattler (`ensure_tool`/`install_tool`, both gated out) and resolves binaries from `~/.ana`, and there is no conda package for it to fall back to. Every invocation would fail in a conda build, so the command doesn't exist there.
- **`self-replace`/`update` module** stay compiled-in under `allow(dead_code)` — harmless, and keeps the gating simple.

**Nit not taken:** merging `anaconda_cli.rs`'s spawn into `run_tool_binary` — the distinct error messages ("Run \`ana bootstrap\` first." vs "Run \`ana tool install\` first.") are deliberate UX.

**New test coverage** (`tests/test_conda_package.py`, 10 tests): builds the package, installs it into a real conda env with the full dependency tree, and verifies version/help, self-update and tool-management unavailability, `tool list` reflecting conda-meta, operation without `CONDA_PREFIX`, and `ana org` proxying to the dep-provided `anaconda` binary.

One finding from the new tests, unrelated to this diff: `anaconda-repo-cli` currently has **no main-label release** on anaconda-cloud (all files are `dev`-labeled). The test works around it with `-c anaconda-cloud/label/dev`, but users won't be able to `conda install anaconda-cli` from the main channel until repo-cli ships a main release.

src/mcp/run.rs
```diff
@@ -1,8 +1,51 @@
...
+    }
+
+    // Look for anaconda-mcp-*.json in conda-meta
+    std::fs::read_dir(&conda_meta)
+        .map(|entries| {
+            entries.filter_map(|e| e.ok()).any(|entry| {
+                entry
+                    .file_name()
+                    .to_string_lossy()
+                    .starts_with("anaconda-mcp-")
```
WebReflection: to be sure we avoid potential directories in there, or unrelated files, shouldn't we also filter `.ends_with(".json")` ?

zfralish: Good catch — fixed in https://github.com/anaconda/anaconda-cli/commit/aba108a. The scan now requires both the `anaconda-mcp-` prefix and the `.json` extension.

WebReflection (approved):

PR #323, opened 2026-07-09
feat: Add conda as a managed tool with wrapper-based activation

zfralish:
> [!NOTE]
> This PR is stacked on #322 (`poc/conda-ship`). Please review and merge #322 first.

## Summary

This PR adds conda as a managed tool in ana, implementing a lightweight standalone ana-managed conda distribution, using conda-spawn for activation.

### How it works

**Installation backends:**

| Feature Flag | Backend | Description |
|--------------|---------|-------------|
| Default (no flag) | Pure rattler | Uses rattler directly to install packages from the lockfile |
| `--features fleet` | conda-ship Fleet API | Uses [conda-ship](https://github.com/jezdez/conda-ship)'s Fleet API for installation |

Both backends produce the same result: a conda installation in `~/.ana/tools/conda/` with the wrapper binary in `~/.ana/bin/conda`.

**Wrapper-based architecture:**

Instead of symlinking to the actual conda binary, we install a standalone wrapper binary (`src/wrappers/conda.rs`) that:
- Intercepts `activate`, `deactivate`, and `init` commands with helpful messages pointing to conda-spawn
- Aliases `conda shell <env>` → `conda spawn <env>` for subshell-based activation
- Filters `create` output to show spawn instructions
- Shows feedback hint on errors
- Passes through all other commands to the real conda

**Post-install configuration:**
- Writes `.condarc` with Anaconda channel defaults
- Creates frozen marker (CEP 22) to prevent accidental base environment modification

### Usage

```bash
# Install conda
ana tool install conda

# Create an environment
conda create -n myenv python=3.12 -y

# Activate (subshell-based)
conda shell myenv

# Exit the environment
exit
```

### Changes

- `build.rs` - Compile standalone conda wrapper binary at build time
- `src/wrappers/conda.rs` - Standalone wrapper that intercepts/aliases commands
- `src/tools/specs.rs` - Add conda tool spec with `uses_wrapper` field
- `src/tools/install.rs` - Add wrapper installation, conda config, frozen marker (rattler backend)
- `src/tools/fleet.rs` - Add wrapper installation, conda config, frozen marker (fleet backend)
- `tool-specs/conda/` - Lockfile, pixi.toml, and .condarc configuration
- `tests/test_tools.py` - Comprehensive tests for conda installation and wrapper behavior

## Test plan

- [x] `cargo clippy` passes
- [x] `cargo test` passes (261 tests)
- [x] `pixi run pytest tests/test_tools.py -k conda` passes (11 tests)
- [x] Manual test: `ana tool install conda` installs successfully
- [ ] Manual test: `conda --version` works via wrapper
- [ ] Manual test: `conda activate` shows helpful error
- [ ] Manual test: `conda shell myenv` spawns subshell
- [ ] Manual test: `conda create -n test python -y` works
- [ ] Manual test with fleet: `cargo run --features fleet -- tool install conda`


src/tools/fleet.rs
```diff
@@ -91,7 +91,14 @@ pub async fn install_tool(ctx: &mut CommandContext, name: &str) -> miette::Resul
...
     );
 
-    create_bin_symlinks(&installed, &binaries)?;
+    let uses_wrapper = specs::uses_wrapper(name);
+    create_bin_symlinks(&installed, &binaries, uses_wrapper)?;
+
+    // For conda, write config and frozen marker
+    if name == "conda" {
+        write_conda_config(&installed.prefix)?;
+        write_frozen_marker(&installed.prefix)?;
```
mattkram: We need to determine whether this should get passed into fleet APIs, but that might be a follow-up action as we refine the approach.

src/tools/install.rs
```diff
@@ -218,19 +243,72 @@ pub async fn install_from_lockfile(
 }
 
 /// Create symlinks (Unix) or shims (Windows) for the tool's binaries in ~/.ana/bin/
-fn create_bin_symlinks(prefix: &Path, binaries: &[PathBuf]) -> miette::Result<()> {
+fn create_bin_symlinks(
+    prefix: &Path,
+    binaries: &[PathBuf],
+    uses_wrapper: bool,
+) -> miette::Result<()> {
```
mattkram: This and the .condarc copy part feel identical between fleet and our existing approach. Can you please check, and refactor as appropriate if it makes sense?

Of course, we might drop the feature flag anyway, which will render it moot, but I'd prefer to not duplicate if we can help it.

src/tools/specs.rs
```diff
@@ -30,13 +34,27 @@ const TOOLS: &[Tool] = &[
...
+        auto_update: true,
+    },
+    Tool {
+        name: "conda",
+        lockfile: include_str!("../../tool-specs/conda/pixi.lock"),
+        binaries: if cfg![unix] {
+            &[&["bin", "conda"]]
+        } else {
+            &[&["Scripts", "conda"]]
+        },
```
mattkram, line 48: Can you double check if we need this? With a wrapper binary, do we actually need to expose the actual binary with the generated shim or symlink?

It's likely the current state is it creates a default wrapper, and then immediately overwrites it.

src/tools/specs.rs
```diff
@@ -9,6 +9,9 @@ struct Tool {
     binaries: &'static [&'static [&'static str]],
     /// If set, the tool is experimental and this message will be shown as a warning.
     experimental: Option<&'static str>,
+    /// If true, the binary in ~/.ana/bin/ should be a link to ana itself,
+    /// which acts as a wrapper for the actual tool binary.
```
mattkram: At one point, ana itself was the wrapper, but IIRC our wrapper binary is its own standalone embedded executable.

src/tools/specs.rs
```diff
@@ -83,6 +101,11 @@ pub fn experimental_message(name: &str) -> Option<&'static str> {
     find_tool(name).and_then(|t| t.experimental)
 }
 
+/// Returns whether a tool uses a wrapper binary (ana acts as the binary).
```
mattkram:
```suggestion
/// Returns whether a tool uses a custom wrapper binary.
```

src/wrappers/conda.rs
```diff
@@ -0,0 +1,351 @@
+//! Standalone conda wrapper binary for ana.
+//!
+//! This binary is compiled separately and embedded into ana, then written to
+//! ~/.ana/bin/conda when `ana tool install conda` is run.
+//!
+//! The wrapper:
+//! - Intercepts `activate`, `deactivate`, and `init` commands with helpful messaging
+//! - Aliases `shell` -> `spawn` for conda-express compatibility
```
mattkram:
```suggestion
//! - Aliases `shell` -> `spawn`
```

We need to decide if we want to do this. Personally, I prefer the API, but we should discuss w/ conda team about what envisioned future general command is.

src/wrappers/conda.rs
```diff
@@ -0,0 +1,351 @@
...
+    format!("\x1b[32m{}\x1b[0m", s)
+}
+
+fn cyan(s: &str) -> String {
+    format!("\x1b[36m{}\x1b[0m", s)
+}
+
+fn blue(s: &str) -> String {
+    format!("\x1b[34m{}\x1b[0m", s)
+}
```
mattkram, line 67: This is probably fine, and it could be YAGNI, but potentially we would want to consider a core refactor where we extract some stuff from ana into a crate and then list that as an internal mutual dependency. 

For now, the amount of duplication is probably not worth the effort, but the paths in particular are slightly concerning in terms of maintainability and consistency.

src/wrappers/conda.rs
```diff
@@ -0,0 +1,351 @@
...
+    eprintln!(
+        "  Just add {} to your PATH and you're ready to go.",
+        cyan("~/.ana/bin")
+    );
+    eprintln!();
+    eprintln!("  To activate environments, use:");
+    eprintln!();
+    eprintln!("    {}", green("conda shell myenv"));
+    eprintln!();
+    eprintln!("  Learn more: https://github.com/conda-incubator/conda-spawn");
```
mattkram: TODO: let's decide what to print for these. We might want to document ourselves and direct there.

src/wrappers/conda.rs
```diff
@@ -0,0 +1,351 @@
...
+            return Some(name.to_string());
+        }
+        if let Some(name) = arg.strip_prefix("--name=") {
+            return Some(name.to_string());
+        }
+    }
+    None
+}
+
+fn run_conda_filtered(args: &[String]) -> i32 {
```
mattkram: I need to take time to think through this more. It's likely we will want to use fleet to handle "run this command in a managed environment" for us. There are a lot of environment variables and such to be careful about, and we need to minimize implementation of conda-specific details as much as we can get away with.

src/wrappers/conda.rs
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    let prefix = conda_prefix();
+    let bin_dir = bin_dir();
+
+    let mut cmd = Command::new(&conda_bin);
+    cmd.args(args);
+    cmd.env("CONDA_ROOT_PREFIX", &prefix);
+
+    // Add ana's bin directory to PATH so our wrapper takes precedence for subcommands
```
mattkram: I'm not sure if this comment is correct, or what it really means.

The wrapper should invoke the `~/.ana/tools/conda/condabin/conda` python entrypoint in a subprocess. I'm not sure where the wrapper would invoke itself ...

tool-specs/conda/.condarc
```diff
@@ -0,0 +1,25 @@
...
+
+# Don't auto-activate base - users should use 'conda shell <env>'
+auto_activate_base: false
+
+# Disable outdated conda notifications (ana handles updates)
+notify_outdated_conda: false
+
+# Packages bundled with ana that should survive 'conda self reset'
+# Note: conda and conda-self are already hardcoded as permanent
+# Note: This setting requires conda-self > 0.1.1 (not yet released)
```
mattkram: Looks like 0.2.0 is on main now.

src/tools/fleet.rs
```diff
@@ -0,0 +1,299 @@
...
+        eprintln!();
+    }
+
+    let lock_content =
+        specs::content(name).ok_or_else(|| miette::miette!("unknown tool: {}", name))?;
+
+    let binaries = specs::binaries(name).unwrap_or_default();
+    let binary_names = specs::binary_names(name).unwrap_or_default();
+
+    let delegate = binary_names.first().copied().unwrap_or(name);
```
jezdez: Could we give each tool an explicit delegate executable, separately from the binaries exposed on PATH? For `anaconda-cli`, the exposed list is intentionally empty, so this selects `anaconda-cli`, but the installed command is `anaconda`, as [the MCP caller already uses](https://github.com/anaconda/anaconda-cli/blob/9bd8d50efa39780f6400beae14f38bd5e8f8601a/src/mcp/run.rs#L45-L53). Fleet checks that the delegate exists before marking installation complete, so a fresh Fleet install of this tool would fail after installing its packages.

The neighboring `requested_specs` field should come from the requested conda packages too. Fleet records those package requests in `conda-meta/history`, and executable names do not necessarily match package names.

src/tools/mod.rs
```diff
@@ -1,17 +1,89 @@
...
+#[cfg(all(tool_install, not(feature = "fleet")))]
+pub async fn ensure_tool(ctx: &mut CommandContext, name: &str) -> miette::Result<()> {
+    install::ensure_tool(ctx, name).await?;
+    Ok(())
+}
+
+/// Ensure a tool is installed, installing it if necessary (fleet version).
+#[cfg(all(tool_install, feature = "fleet"))]
+pub async fn ensure_tool(ctx: &mut CommandContext, name: &str) -> miette::Result<()> {
+    if !crate::paths::tool_prefix(name).exists() {
```
jezdez: An existing directory can contain an interrupted installation or an older version of a tool. This check skips both. One detail of [Fleet 0.9.1](https://github.com/conda-incubator/conda-ship/blob/71788f3451524a4859ddc6435067cb176919832c/src/fleet.rs#L61-L69) is that `install` with default options also reuses a ready installation without comparing the new lockfile.

Could we compare a healthy runtime's `lock_sha256` with `RuntimeSpec::lock_sha256()` and use `force: true` when an update is selected? Interrupted installs should still reach `Fleet::install` so it can recover them. The same logic could support `tool update` and the post-self-update path, keeping ana's existing auto-update preferences. A forced Fleet reinstall preserves named environments and unrelated files.

src/tools/fleet.rs
```diff
@@ -0,0 +1,299 @@
...
+
+    let spec = RuntimeSpec {
+        id: name.to_string(),
+        version: tool_version_from_lock(&lock_content, name)?,
+        delegate_executable: delegate.to_string(),
+        lock_content,
+        requested_specs,
+        // .condarc and the frozen marker are written post-install below
+        condarc: None,
+        freeze_base: false,
```
jezdez: Fleet 0.9.1 already supports both of these options, so we can pass the conda `.condarc` text and `freeze_base: true` here. Fleet [writes them before marking the prefix ready](https://github.com/conda-incubator/conda-ship/blob/71788f3451524a4859ddc6435067cb176919832c/src/fleet.rs#L115-L148). With the separate writes below, a failure can leave a ready installation and an installed wrapper without the intended configuration or base protection.

Fleet currently uses a generic frozen-environment message. If we want to preserve ana's specific guidance, a configurable message would be a useful small addition to conda-ship.

src/tools/fleet.rs
```diff
@@ -0,0 +1,299 @@
...
+/// Check if a prefix is an old rattler-based installation (pre-Fleet).
+///
+/// Old installations have:
+/// - A `conda-meta/` directory (bootstrapped conda prefix)
+/// - A `.lockfile-hash` file (ana's old staleness marker)
+/// - No `.{name}.json` Fleet metadata file
+fn is_legacy_rattler_install(prefix: &Path, name: &str) -> bool {
+    let conda_meta = prefix.join("conda-meta");
+    let lockfile_hash = prefix.join(".lockfile-hash");
+    let fleet_metadata = conda_meta.join(format!(".{}.json", name));
```
jezdez: Small path correction: Fleet puts its metadata at `<prefix>/.<name>.json`, directly under the prefix, as shown in [the Fleet layout](https://github.com/conda-incubator/conda-ship/blob/71788f3451524a4859ddc6435067cb176919832c/docs/explanation/fleet.md#L12-L24). Could we build this path from `prefix` instead of `conda_meta`? If the old `.lockfile-hash` exists alongside valid Fleet metadata, the current check classifies that Fleet installation as legacy and sends it into the deletion path below.

src/tools/fleet.rs
```diff
@@ -0,0 +1,299 @@
...
+/// Migrate a legacy rattler-based installation to Fleet.
+///
+/// This removes the old prefix so Fleet can do a fresh install.
+fn migrate_legacy_install(prefix: &Path, name: &str) -> miette::Result<()> {
+    crate::ui::status::info(&format!(
+        "Migrating {} from legacy installation to Fleet...",
+        name
+    ));
+
+    std::fs::remove_dir_all(prefix)
```
jezdez: There is a conda-specific detail to account for here: a writable conda installation normally stores named environments under `<prefix>/envs`. For example, someone using the default backend could create an environment and then switch to a Fleet build. This recursive removal would delete their environment along with conda. Freezing base does not prevent users from creating those child environments.

Could we preserve user environments during migration? For this first version, stopping with an explanation when they are present would be a reasonable starting point. Fleet's `force` option cannot adopt a legacy prefix, so it would not be a replacement for this migration step.

jezdez (commented):
Thanks for putting this together, Zeke. Fleet is a good fit for ana managing several tools from embedded lockfiles. ana can keep ownership of tool selection, wrappers, prompts, and update policy, while Fleet handles package installation, recovery, and installed-prefix metadata.

I've left a few comments explaining the Fleet behavior that matters here, with suggestions for the adapter. The main things to address are selecting the right executable, handling updates and interrupted installs, and preserving user environments during migration. Fleet's documentation could also be clearer about how it handles an already-installed runtime.

On the wrapper question raised earlier, I think `uses_wrapper` should stay in ana. Fleet's [command and shim APIs](https://github.com/conda-incubator/conda-ship/blob/71788f3451524a4859ddc6435067cb176919832c/docs/reference/fleet.md#L143-L172) can supply executable paths and PATH entries, while ana still launches the process and handles feedback. Those APIs do not perform shell activation.

It would also be useful for CI to exercise the tool integration tests with the Fleet backend, since the current jobs use the default backend.

src/tools/fleet.rs
zfralish: Fixed in 8d2ca3e: https://github.com/anaconda/ana-cli/commit/8d2ca3e

Went with a `delegate` field on the tool spec (set to `"anaconda"` for anaconda-cli, defaulting to the tool name for everything else) rather than a `match`, so the catalog keeps all per-tool metadata in one place. `requested_specs` now records the package name instead of binary names.

src/tools/mod.rs
zfralish: Addressed in https://github.com/anaconda/ana-cli/commit/fed3b19 — applied both suggested edits (`ensure_tool` hash check, `force` on hash mismatch in `install_tool`), with `tool update` and the post-self-update path now using the same logic under the existing auto-update preferences. One deviation: sha2 0.11's digest has no `LowerHex` impl, so hex encoding lives in a small `fleet::lock_hash` helper. Delegate-missing repair left as a follow-up per your scoping.

src/tools/fleet.rs
zfralish: Done in https://github.com/anaconda/ana-cli/commit/93dfa91 — condarc and freeze_base now go through the spec, the post-install block is removed, and the common.rs helpers are gated `not(feature = "fleet")`. Verified with a fleet build: `.condarc` and the frozen marker (Fleet's message) are written before the prefix is marked ready, and the wrapper works. The frozen-marker test now accepts either backend's wording. +1 on a configurable frozen message in conda-ship as a follow-up to restore ana's `conda self` guidance.

src/tools/fleet.rs
zfralish: Fixed in https://github.com/anaconda/ana-cli/commit/1882063 — metadata path now built from `prefix`. Also confirmed empirically from a fleet install that the file lands at `<prefix>/.conda.json`.

src/tools/fleet.rs
zfralish: Applied in https://github.com/anaconda/ana-cli/commit/b4f2d30 — the guard runs before the "Migrating..." message so a refusal doesn't print a misleading notice. Added unit tests covering non-empty envs (refused, prefix preserved), empty envs (removed), and no envs (removed). Agreed a broader migration would need to account for other user files; leaving that for later.

zfralish:
All review comments addressed — one commit each, linked in the threads.

On the summary points: CI now exercises the Fleet backend in https://github.com/anaconda/ana-cli/commit/2628b1c — a dedicated step builds with `--features fleet` and runs the tool install/list/update/uninstall + conda wrapper tests against that binary (via `ANA_BINARY_PATH`), without duplicating the full suite.

Running the suite against the fleet backend surfaced three gaps, fixed in that commit:

1. **Reinstalling current tools** — fleet `tool install` always printed "Installing..." and went through Fleet even when the runtime was already current; it now short-circuits with "already up to date" on a hash match (rattler parity).
2. **Broken-symlink cleanup missing from `tool list`** — the rattler backend's `list` removes dangling symlinks in ~/.ana/bin (CLI-526); the fleet `list` didn't. Now both do.
3. **Hash-storage assumption in the update test** — `test_tool_update_updates_installed_tool` corrupted `.lockfile-hash`, which only the rattler backend writes; the test now corrupts `lock_sha256` in the Fleet metadata when running under fleet.

Known remaining gap (out of scope per review discussion): a healthy-metadata install whose delegate executable was deleted is not detected or repaired by `ensure_tool` — flagged as follow-up.

And thanks for the pointer on the command/shim APIs — agreed, `uses_wrapper` stays in ana, which keeps process launching and feedback handling on our side.

jezdez (commented):
Thanks for working through these, Zeke. The delegate, lock-hash update, configuration, and migration changes address the main concerns from the first pass. Thanks also for correcting the SHA-256 formatting in my earlier suggestion.

I found a few remaining issues around launcher repair, version and requested-package metadata, and feature combinations. I have included code suggestions for those, the related tests, and conda-ship 0.9.2. The six `list.rs` guard suggestions should be applied together, as should the version parser and its tests.

I checked these replacements with focused parser, feature-guard, and launcher-flow tests. I have not run a full ana build or end-to-end installation.

Cargo.toml
```diff
@@ -38,10 +40,11 @@ console = "0.16"
...
 log = "0.4"
 miette = { version = "7.6", features = ["fancy"] }
-rattler = { version = "0.47", default-features = false, features = ["indicatif"], optional = true }
+rattler = { version = "0.48", default-features = false, features = ["indicatif"], optional = true }
 rattler_cache = { version = "0.10", default-features = false, optional = true }
-rattler_conda_types = { version = "0.48", default-features = false, optional = true }
-rattler_lock = { version = "0.31", default-features = false, optional = true }
+rattler_conda_types = { version = "0.50", default-features = false, optional = true }
+rattler_lock = { version = "0.32", default-features = false, optional = true }
+conda-ship = { git = "https://github.com/jezdez/conda-ship", tag = "0.9.1", default-features = false, features = ["fleet", "native-tls"], optional = true }
```
jezdez: Could we use conda-ship 0.9.2 and its current repository URL? The Fleet API is unchanged from 0.9.1. Please regenerate `Cargo.lock` with this change so the source URL and resolved commit stay in sync.

```suggestion
conda-ship = { git = "https://github.com/conda-incubator/conda-ship", tag = "0.9.2", default-features = false, features = ["fleet", "native-tls"], optional = true }
```

src/tools/fleet.rs
```diff
@@ -0,0 +1,414 @@
...
+    let lock_content =
+        specs::content(name).ok_or_else(|| miette::miette!("unknown tool: {}", name))?;
+
+    let binaries = specs::binaries(name).unwrap_or_default();
+
+    // The delegate executable and requested package are independent of the
+    // binaries ana exposes on PATH (e.g. anaconda-cli exposes nothing but
+    // provides `bin/anaconda`).
+    let delegate = specs::delegate_executable(name);
+    let requested_specs: Vec<String> = vec![name.to_string()];
```
jezdez: `anaconda-cli` is the runtime ID, but it is not one of the packages requested by its manifest. Fleet writes these values to `conda-meta/history` and prefix metadata. Can we read the actual requests from the existing manifests so the package names and version constraints stay accurate?

```suggestion
    // The delegate and requested packages are independent of exposed binaries.
    let delegate = specs::delegate_executable(name);
    let manifest = match name {
        "anaconda-cli" => include_str!("../../tool-specs/anaconda-cli/pixi.toml"),
        "conda" => include_str!("../../tool-specs/conda/pixi.toml"),
        "pixi" => include_str!("../../tool-specs/pixi/pixi.toml"),
        "outerbounds" => include_str!("../../tool-specs/outerbounds/pixi.toml"),
        _ => return Err(miette::miette!("unknown tool: {name}")),
    };
    let manifest: toml::Value = toml::from_str(manifest)
        .into_diagnostic()
        .context("failed to parse tool manifest")?;
    let requested_specs = manifest
        .get("dependencies")
        .and_then(toml::Value::as_table)
        .ok_or_else(|| miette::miette!("tool manifest has no dependencies"))?
        .iter()
        .map(|(package, version)| {
            let version = version.as_str().ok_or_else(|| {
                miette::miette!("expected a version string for dependency {package}")
            })?;
            Ok(format!("{package} {version}"))
        })
        .collect::<miette::Result<Vec<_>>>()?;
```

src/tools/fleet.rs
```diff
@@ -0,0 +1,414 @@
...
+        )
+        .await
+        .with_context(|| format!("failed to install tool: {}", name))?;
+
+    eprintln!(
+        "   Installed {} v{} to {}",
+        installed.id,
+        installed.version,
+        installed.prefix.display()
+    );
```
jezdez: Could we reuse the current runtime here and still run the launcher setup below? Fleet marks the prefix ready before ana writes its wrappers and shims. If that write fails, or a wrapper is later deleted, the early return prevents `ana tool install conda` from repairing it. This also lets the existing pixi configuration step run on retry.

```suggestion
    // Reuse current prefixes while still refreshing ana's launchers below.
    let desired_hash = spec.lock_sha256();
    let existing = fleet.get(name)?;
    let force = existing.is_some();
    let installed = match existing {
        Some(runtime) if runtime.lock_sha256.as_deref() == Some(desired_hash.as_str()) => {
            eprintln!("{} is already up to date.", name);
            runtime
        }
        _ => {
            eprintln!("Installing {} into {}", name, prefix.display());

            let installed = fleet
                .install(
                    spec,
                    InstallOptions {
                        force,
                        ..InstallOptions::default()
                    },
                )
                .await
                .with_context(|| format!("failed to install tool: {}", name))?;

            eprintln!(
                "   Installed {} v{} to {}",
                installed.id,
                installed.version,
                installed.prefix.display()
            );
            installed
        }
    };
```

src/tools/fleet.rs
```diff
@@ -0,0 +1,414 @@
...
+                if parts.len() >= 2 {
+                    return Ok(parts[1].to_string());
+                }
+            }
+        }
+    }
+
+    // Fallback to "latest" if we can't extract a version
+    Ok("latest".to_string())
+}
```
jezdez: The version parser still records `cli` for `anaconda-cli`, because it matches `anaconda-cli-base-0.9.1` and takes the second hyphen-separated component. Could we use the structured lock records for the current platform and select `anaconda-cli-base` explicitly? The existing rattler dependencies already provide the parser.

```suggestion
/// Extract the tool package version from the lockfile for this platform.
fn tool_version_from_lock(lock_content: &str, tool_name: &str) -> miette::Result<String> {
    let package_name = match tool_name {
        "anaconda-cli" => "anaconda-cli-base",
        name => name,
    };
    let lock_file = rattler_lock::LockFile::from_str_with_base_directory(lock_content, None)
        .into_diagnostic()
        .context("failed to parse lockfile")?;
    let environment = lock_file
        .default_environment()
        .ok_or_else(|| miette::miette!("lockfile has no default environment"))?;
    let platform = rattler_conda_types::Platform::current();
    let records = environment
        .conda_repodata_records_by_platform()
        .into_diagnostic()
        .context("failed to extract records from lockfile")?
        .into_iter()
        .find(|(p, _)| p.subdir() == platform)
        .map(|(_, records)| records)
        .ok_or_else(|| miette::miette!("lockfile has no records for platform {platform}"))?;

    records
        .into_iter()
        .find(|record| record.package_record.name.as_normalized() == package_name)
        .map(|record| record.package_record.version.to_string())
        .ok_or_else(|| {
            miette::miette!("lockfile has no {package_name} package for platform {platform}")
        })
}
```

src/tools/fleet.rs
```diff
@@ -0,0 +1,414 @@
...
+        let version = tool_version_from_lock(lock_content, "pixi").unwrap();
+        assert_eq!(version, "0.70.2");
+    }
+
+    #[test]
+    fn test_tool_version_from_lock_fallback() {
+        let lock_content = "version: 6\n";
+        let version = tool_version_from_lock(lock_content, "unknown").unwrap();
+        assert_eq!(version, "latest");
+    }
```
jezdez: Along with the structured parser, these fixtures need valid lockfile package records. The embedded `anaconda-cli` lock also covers the hyphenated package name that currently produces `cli`.

```suggestion
    #[test]
    fn test_tool_version_from_lock() {
        for (name, lock_content, expected) in [
            (
                "pixi",
                include_str!("../../tool-specs/pixi/pixi.lock"),
                "0.70.2",
            ),
            (
                "anaconda-cli",
                include_str!("../../tool-specs/anaconda-cli/pixi.lock"),
                "0.9.1",
            ),
        ] {
            assert_eq!(
                tool_version_from_lock(lock_content, name).unwrap(),
                expected
            );
        }
    }

    #[test]
    fn test_tool_version_from_lock_missing_package() {
        let lock_content = include_str!("../../tool-specs/pixi/pixi.lock");
        assert!(tool_version_from_lock(lock_content, "unknown").is_err());
    }

    #[test]
    fn test_tool_version_from_lock_invalid() {
        assert!(tool_version_from_lock("version: 6\n", "pixi").is_err());
    }
```

src/tools/list.rs
```diff
@@ -12,6 +12,8 @@ use super::specs;
 pub struct ToolInfo {
     pub name: &'static str,
     pub installed: bool,
+    #[cfg(feature = "fleet")]
```
jezdez: The `version` field needs the same condition as the Fleet listing function below. These six guard suggestions go together, so the existing conda-package listing still works when tool installation is disabled.

```suggestion
    #[cfg(all(tool_install, feature = "fleet"))]
```

src/tools/list.rs
```diff
@@ -46,7 +48,7 @@ const INSTALLERS: &[Installer] = &[
 ];
 
 /// Check if a tool is installed as a conda package in the current environment.
-#[cfg(not(tool_install))]
+#[cfg(all(not(tool_install), not(feature = "fleet")))]
```
jezdez: Please keep this helper available whenever tool installation is disabled, including builds that also enable `fleet`. Those builds still read installation status from the active conda environment.

```suggestion
#[cfg(not(tool_install))]
```

src/tools/list.rs
```diff
@@ -69,9 +71,10 @@ fn is_conda_package_installed(name: &str) -> bool {
 ///
 /// When built with `conda-package` feature, installation status reflects the
 /// conda environment's conda-meta entries rather than `~/.ana/tools/`.
+#[cfg(not(feature = "fleet"))]
```
jezdez: This keeps the existing listing implementation available when Fleet cannot be used because `tool_install` is disabled.

```suggestion
#[cfg(not(all(tool_install, feature = "fleet")))]
```

src/tools/list.rs
```diff
@@ -92,7 +95,44 @@ pub fn list_tools() -> Vec<ToolInfo> {
         .collect()
 }
 
+/// List all available tools with their installation status (fleet version).
+#[cfg(feature = "fleet")]
```
jezdez: Could we match the condition on `mod fleet` here? `--no-default-features --features fleet` and `--all-features` currently compile this function while omitting `super::fleet`, which causes an unresolved-module error. The other five guard suggestions make the field, helper, and renderers consistent with this choice.

```suggestion
#[cfg(all(tool_install, feature = "fleet"))]
```

src/tools/list.rs
```diff
@@ -92,7 +95,44 @@ pub fn list_tools() -> Vec<ToolInfo> {
...
+                installed,
+                version,
+                binaries,
+            }
+        })
+        .collect()
+}
+
 /// Print the tool list as a formatted table.
+#[cfg(not(feature = "fleet"))]
```
jezdez: The existing renderer should follow the same condition as the existing listing function.

```suggestion
#[cfg(not(all(tool_install, feature = "fleet")))]
```

src/tools/list.rs
```diff
@@ -118,6 +158,39 @@ pub fn print_tool_list(_ctx: &mut CommandContext) {
     print_installer_list();
 }
 
+/// Print the tool list as a formatted table (fleet version with version column).
+#[cfg(feature = "fleet")]
```
jezdez: The Fleet renderer also needs `tool_install`, matching the Fleet listing function and its `version` field.

```suggestion
#[cfg(all(tool_install, feature = "fleet"))]
```

tests/test_tools.py
```diff
@@ -388,3 +395,250 @@ def test_tool_uninstall_shows_what_will_be_removed(
...
+        assert result.returncode == 0
+        assert "Installed wrapper" in result.stderr
+
+        # Verify the wrapper binary was created
+        wrapper_name = "conda.exe" if IS_WINDOWS else "conda"
+        bin_path = fake_home / ".ana" / "bin" / wrapper_name
+        assert bin_path.exists(), f"Wrapper binary not found: {bin_path}"
+        # The wrapper is a standalone binary, not a symlink
+        assert not bin_path.is_symlink(), "Wrapper should be a binary, not a symlink"
+        assert bin_path.stat().st_size > 0, "Wrapper binary should not be empty"
```
jezdez: Could we extend this test to cover restoring a missing wrapper? It reuses the installation above and exercises the early-return fix without creating another conda environment.

```suggestion
        assert bin_path.stat().st_size > 0, "Wrapper binary should not be empty"

        bin_path.unlink()
        result = run_ana("tool", "install", "conda")
        assert result.returncode == 0, result.stderr
        assert bin_path.exists(), "Reinstall should restore the missing wrapper"
        assert not bin_path.is_symlink()
        assert bin_path.stat().st_size > 0
```

.github/workflows/ci.yaml
```diff
@@ -169,6 +169,16 @@ jobs:
...
         run: pixi run test-integration
 
+      # Exercise the fleet backend's core flows (install/list/update/uninstall,
+      # conda wrapper) without duplicating the full integration suite.
+      - name: Run fleet backend tool tests
+        if: ${{ !inputs.skip-tests }}
+        env:
+          ANA_BINARY_PATH: ${{ github.workspace }}/target/debug/${{ runner.os == 'Windows' && 'ana.exe' || 'ana' }}
+        run: |
+          pixi run build-fleet
```
jezdez: Could we run the Fleet Rust tests here too? The new tests that preserve user environments during migration are inside the Fleet module, but the existing Rust test commands build without Fleet. Building the binary and running pytest does not execute those Rust tests. This uses the existing version wrapper and keeps the current Python coverage.

```suggestion
          pixi run build-fleet
          pixi run python scripts/with_version.py cargo test --locked --features fleet
```

zfralish:
Addressed in 9afd38d — remaining Fleet review feedback applied (conda-ship 0.9.2, requested specs from manifests, launcher repair on reinstall, structured lock version parsing, list.rs feature guards, wrapper restore test, and Fleet Rust tests in CI).

.github/workflows/ci.yaml
zfralish: I am including fleet as a default feature, so this should just be picked up naturally.

PR #324, opened 2026-07-10
refac: Use owo-colors for terminal color output

zfralish:
## Summary
- Switch from hardcoded hex colors (via `console` crate) to ANSI color codes using `owo-colors`
- Terminal now applies the user's configured color profile instead of forcing specific RGB values
- Keeps `console` crate for terminal utilities (`Term`, `Key`, `strip_ansi_codes`)

CleanShot 2026-07-10 at 10 36 19

## Test plan
- [ ] Run `ana --help` and verify colors display correctly
- [ ] Run `ana auth status` and verify colored output
- [ ] Test with `NO_COLOR=1 ana --help` to verify colors are disabled
- [ ] Test in a terminal with custom color scheme to verify user's colors are respected


mattkram (approved):
LGTM. I think we have an existing bug in the pre-commit job. If you run it locally and commit the change it's trying to auto-commit I think we can make CI green.

PR #325, opened 2026-07-10
ci: Use JWT to access the HashiCorp Vault

marcoesters:
## Summary

Use JWT to access the HashiCorp Vault using a short-lived token obtained via OIDC. This reduces the need for secrets that need to be managed and rotated if a leak were to occur.

Access is controlled via policy, so only authorized repositories can access the Vault.


mattkram (approved):
Looks great!

PR #330, opened 2026-07-14
chore: Add Slack notification for releases

zfralish:
## Summary
- Adds a new `notify-slack` job to the release workflow that posts release notes to a Slack channel when a GitHub release is published
- Uses Slack Block Kit for rich formatting with release title, notes, and a button to view the release

## Test plan
- [ ] Add `SLACK_RELEASE_WEBHOOK_URL` secret to repository settings
- [ ] Create a test release and verify Slack notification is received
- [ ] Verify message contains release version, notes, and working link


.github/workflows/release.yaml
```diff
@@ -95,6 +95,56 @@ jobs:
           echo "Publishing release $VERSION"
           gh release edit "$VERSION" --draft=false --latest
 
+  notify-slack:
```
mattkram, line 137: I'm wondering if maybe we want to move this after the "deploy" step below. The self-update feature operates off of anaconda.sh, so users won't be able to update until that is completed.

.github/workflows/release.yaml
```diff
@@ -95,6 +95,56 @@ jobs:
...
+                    "type": "header",
+                    "text": {
+                      "type": "plain_text",
+                      "text": "🚀 Ana CLI \($tag) Released"
+                    }
+                  },
+                  {
+                    "type": "section",
+                    "text": {
+                      "type": "mrkdwn",
```
mattkram, line 167: Why would they do this to us?

mattkram (approved):
A few small comments, but looks good. Can you please also adjust the type in the PR title to "chore" since its an internal-only thing?

.github/workflows/release.yaml
zfralish: I guess vowels are expensive or something

PR #332, opened 2026-07-15
chore: Enable pre-commit auto-push and resolve zizmor warnings

zfralish:
## Summary
- Remove `persist-credentials: false` from pre-commit workflow so the bot token can push auto-fix commits
- Fix zizmor template injection warnings in release.yaml Slack notification by moving `github.event.release` values to env vars

This fixes the failing pre-commit CI on all Renovate PRs.

## Test plan
- [ ] Verify pre-commit workflow can push commits on a PR branch
- [ ] Verify zizmor passes locally and in CI


mattkram (commented):
Is there a way we could test this by making an intentional formatting change that should get fixed?

mattkram (approved):
Awesome, thanks!

PR #343, opened 2026-07-24
feat: Display release notes after self update

zfralish:
## Summary
Fetches and displays structured release notes from anaconda.sh after a successful `ana self update`, showing users what's new in the version they just installed.

- Adds `src/release_notes.rs` module for fetching and formatting release notes
- Integrates into the update flow to display after successful updates
- Shows new features (highlighted) and bug fixes (dimmed) in a "WHAT'S NEW" section
- Strips conventional commit prefixes for cleaner display
- Silently handles fetch failures (doesn't break update flow)

Example output after update:

CleanShot 2026-07-24 at 09 32 01

CleanShot 2026-07-24 at 09 32 37

## Test plan
- [ ] Run `ana self update` when a new version is available
- [ ] Verify release notes display after successful update
- [ ] Test with version that has no release notes (should complete silently)
- [ ] Run `cargo test release_notes` to verify unit tests pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -246,6 +246,7 @@ dependencies = [
  "sha2 0.11.0",
  "temp-env",
  "tempfile",
+ "termimad",
  "thiserror 2.0.19",
  "tokio",
  "toml 1.1.3+spec-1.1.0",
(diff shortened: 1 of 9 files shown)


mattkram (commented):
Love where you are going w/ this.

A few design notes before reviewing the code:

* We probably don't need to show maintenance stuff
* Let's show a link to the release on GitHub, instead of the default "compare tags" link that GitHub provides. Curious users can dig in from the release
* We will need to decide how we aggregate changes if the user upgrades several versions at once
* Let's make each entry minimal, and aligned. We don't need to show the "by author", or probably the link to the PR. Maybe just the PR number? Check out the below from `omz changelog` for inspiration. We don't need to match exactly, but I think some tab alignment and reduction in characters will help us here.

Image

PR #344, opened 2026-07-27
fix: Use default_channels instead of channels for main-x feature

zfralish:
## Summary
- Changed `ana feature enable main-x` to add main-x to `default_channels` instead of `channels` in `.condarc`
- This ensures main-x is resolved as part of the defaults multi-channel, producing correct channel priority behavior

## Test plan
- [x] Run `ana feature enable main-x` and verify the channel is added to `default_channels` (not `channels`) in `.condarc`
- [x] Run `ana feature disable main-x` and verify the channel is removed from `default_channels`
- [x] Verify conda can resolve packages from main-x correctly with this configuration

### DIFF
--- a/src/feature/main_x.rs
+++ b/src/feature/main_x.rs
@@ -786,9 +786,11 @@ mod tests {
         // Should add all 4 required default_channels plus "defaults" to channels
         assert_eq!(actions.len(), 5);
         // Check that defaults is added to channels
-        assert!(actions
-            .iter()
-            .any(|a| matches!(a, MainXCondaAction::EnsureDefaultsInChannels)));
+        assert!(
+            actions
+                .iter()
+                .any(|a| matches!(a, MainXCondaAction::EnsureDefaultsInChannels))
+        );
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


src/feature/main_x.rs
```diff
@@ -27,25 +27,25 @@ enum CondaChannelAction {
 }
 
 impl CondaChannelAction {
-    fn commands(&self) -> Vec<(&'static str, &'static str)> {
+    fn commands(&self) -> Vec<(&'static str, &'static str, &'static str)> {
         match self {
             CondaChannelAction::AddMain => {
-                vec![("--add", MAIN_CHANNEL)]
+                vec![("--add", "default_channels", MAIN_CHANNEL)]
```
zfralish: @mattkram should this be left where it was? Or should it also be added to `default_channels`?

condathan (approved):
Tested using debug build and works. The only note I have is enabling main-x will potentially add 3 other channels, but disabling main-x will keep those channels. A user might be surprised to see other Anaconda channels still enabled. Maybe a message after disabling could be printed to be transparent.

SaundersR1999 (approved):
tested, lgtm

redeux (approved):
Approved

PR #366, opened 2026-08-06
fix: Detect corrupted tool installation when binary is missing (CLI-678)

Umanan23:
When a tool's binary is manually deleted or corrupted, `ana tool install` was reporting 'Already up to date' instead of detecting the broken state.

This fix adds a `binaries_exist()` check in `ensure_tool()` that verifies all expected binaries are present in the prefix directory. If the prefix exists and the lockfile hash matches but binaries are missing, the corrupted installation is removed and a clean reinstall is performed.

Also updates `installed_tools()` and `list_tools()` to check that binaries actually exist, so `ana tool list` accurately reflects broken/missing state.

Fixes CLI-678

### DIFF
--- a/src/tools/install.rs
+++ b/src/tools/install.rs
@@ -85,7 +85,12 @@ pub async fn ensure_tool(ctx: &mut CommandContext, name: &str) -> miette::Result
     if prefix.exists() {
         std::fs::remove_dir_all(&prefix)
             .into_diagnostic()
-            .with_context(|| format!("failed to remove corrupted tool directory: {}", prefix.display()))?;
+            .with_context(|| {
+                format!(
+                    "failed to remove corrupted tool directory: {}",
+                    prefix.display()
+                )
+            })?;
     }
 
     install_tool(ctx, name).await?;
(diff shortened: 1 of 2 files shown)


src/tools/install.rs
```diff
@@ -88,6 +105,28 @@ fn hash_lockfile(content: &str) -> String {
...
+        #[cfg(windows)]
+        let path = path.with_extension("exe");
+        path.exists()
+    })
+}
+
+fn validation_binaries(name: &str) -> Option<Vec<PathBuf>> {
+    if name == "anaconda-cli" {
+        let bin_subdir = if cfg!(windows) { "Scripts" } else { "bin" };
+        return Some(vec![PathBuf::from(bin_subdir).join("anaconda")]);
```
mattkram: These should be managed by rattler, and it's not straightforward to know the binaries we are searching for. This one in particular is buried. Either it should live with the tool spec, or more likely we might want to reconsider this fix.

If the symlink/shim is already being cleaned up and the tool bin directory is not on path, in what scenarios will the binaries be missing (with the exception of manual manipulation)?

Umanan23: @mattkram Only via manual deletion/corruption of the prefix contents after a successful install, the prefix `bin/` isn't on PATH, and `cleanup_broken_symlinks` already handles the symlink side. Agreed this shouldn't be a hardcoded list in `install.rs`; I'll move it to check against rattler's own `PrefixRecord` paths instead.

